PR #1, opened 2023-05-29
restore user account

calellowitz:
This adds a user account restore workflow (when the user has a new device and needs to regain access to their account). It validates the user by requiring an OTP from both their primary and secondary phone numbers, and attempts to enforce security throughout the process by generating a secret key on the initial request that is required for all future requests. It also enforces a specific ordering of steps via a database model that keeps the current state. Full API description in https://docs.google.com/spreadsheets/d/1J0tiUK0Hf7LSCxne-4wwQHOykIsl_Hhaq95B6LtwRec/edit#gid=0


users/views.py
```diff
@@ -67,3 +70,127 @@ def confirm_otp(request):
...
+
+
+@require_POST
+@csrf_exempt
+def confirm_recovery_otp(request):
+    data = json.loads(request.body)
+    phone_number = data["phone"]
+    secret_key = data["secret_key"]
+    user = ConnectUser.objects.get(phone_number=phone_number)
+    status = RecoveryStatus.objects.get(user=user)
```
snopoke, line 127: I could see there being multiple of these for a user in an edge case so perhaps better to add the secret key to the DB filter and check for an existing record.

users/views.py
```diff
@@ -67,3 +70,127 @@ def confirm_otp(request):
...
+        return HttpResponse(status=401)
+    if status.step != RecoveryStatus.RecoverySteps.CONFIRM_SECONDARY:
+        return HttpResponse(status=401)
+    device = PhoneDevice.objects.get(phone_number=user.recovery_phone, user=user)
+    verified = device.verify_token(data.get('token'))
+    if not verified:
+        return JsonResponse({"error": "OTP token is incorrect"}, status=401)
+    status.step = RecoveryStatus.RecoverySteps.RESET_PASSWORD
+    status.save()
+    return HttpResponse()
```
snopoke, line 178: maybe not worth it but these primary / secondary views seem like they share a lot of logic that could be extracted.

snopoke (approved):
no blockers

snopoke (approved):
no blockers

PR #2, opened 2023-05-29
Updates to core APIs

calellowitz:
This PR is a bit of a catchall for the rest of the V1 workflows, but includes switching rate limiting and auth to django-rest-framework, switching dependency management to pip-tools, and add recovery options where the user has their password.


snopoke (approved):

users/views.py
```diff
@@ -234,17 +234,17 @@ def phone_available(request):
 @permission_classes([])
 def change_phone(request):
     data = request.data
-    user = ConnectUser.objects.get(phone_number=phone_number)
```
snopoke: 👍🏻

snopoke (approved):

PR #4, opened 2023-07-31
Ce/OIDC

calellowitz:
This PR sets up the OAuth provider and configures a few extra fields to be returned by the openid connect userinfo view.


users/views.py
```diff
@@ -150,17 +146,18 @@ def recover_secondary_phone(request):
         return HttpResponse(status=401)
     if status.step != RecoveryStatus.RecoverySteps.CONFIRM_SECONDARY:
         return HttpResponse(status=401)
-    device = PhoneDevice.objects.get(phone_number=user.recovery_phone, user=user)
-    device.generate_challenge()
+    otp_device, _ = PhoneDevice.objects.get_or_create(phone_number=user.recovery_phone, user=user)
+    otp_device.save()
```
snopoke: what's the extra `save` call for?

users/oauth.py
```diff
@@ -0,0 +1,11 @@
+from oauth2_provider.oauth2_validators import OAuth2Validator
+
+class ConnectOAuth2Validator(OAuth2Validator):
+
+    def get_additional_claims(self, request):
+        claims = {}
+        claims["sub"] = request.user.username
```
snopoke, line 7: what is `sub`?

users/views.py
calellowitz: I think this is a weird merge artifact. I did not change that view, and it's not in the full diff.

users/oauth.py
calellowitz: It stands for "subscriber." It is the normal oauth field for the user identifier (in theory we could use anything but it seemed more correct to just modify this field). It defaults in django-oauth-toolkit to the primary key in the user table, but username seemed more reasonable. You can see it used in the SSO PR when we are saving the link.

snopoke (approved):

PR #5, opened 2023-08-24
add fetch user endpoint

calellowitz:
This adds an API endpoint that accepts list of phone numbers, and returns ConnectID usernames for the users it finds, as well as a list of the numbers it was unable to match. I plan to change the auth to require client credentials (basically restricting it to the CCC server), but wanted to open a PR to get review started.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -264,3 +264,15 @@ def change_password(request):
     user.set_password(password)
     user.save()
     return HttpResponse()
+
+
+@api_view(['GET'])
+def fetch_users(request):
+    data = request.data
+    numbers = data.get('phone_numbers')
+    results = {}
+    found_users = list(User.objects.filter(phone_number__in=numbers).values('username', 'phone_number'))
+    results["found_users"] = found_users
+    missing_numbers = set(numbers) - {u["phone_number"] for u in found_users}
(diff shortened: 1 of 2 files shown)


users/views.py
```diff
@@ -264,3 +264,15 @@ def change_password(request):
     user.set_password(password)
     user.save()
     return HttpResponse()
+
+
+@api_view(['GET'])
```
snopoke, line 269: this should have some auth protection

users/views.py
```diff
@@ -264,3 +264,15 @@ def change_password(request):
...
+def fetch_users(request):
+    data = request.data
+    numbers = data.get('phone_numbers')
+    results = {}
+    found_users = list(User.objects.filter(phone_number__in=numbers).values('username', 'phone_number'))
+    results["found_users"] = found_users
+    missing_numbers = set(numbers) - {u["phone_number"] for u in found_users}
+    results["missing_users"] = missing_users
```
snopoke, line 277: I'd leave this out and let the client figure it out

snopoke (commented):
Generally looks OK to me

snopoke (approved):

snopoke (approved):

PR #6, opened 2023-10-03
alphanumeric sender

calellowitz:
Some Malawi telecoms require alphanumeric senders instead of international numbers. This PR means messages to malawi will come from "ConnectID" instead of the US number we purchased in twilio


pxwxnvermx (approved):
Thanks 🙌🏼

PR #7, opened 2023-10-06
push notifications

snopoke:
* register FCM token during user registration or via heartbeat endpoint
* API to send individual messages or bulk messages

Jira: https://dimagi-dev.atlassian.net/browse/CCPC-133


users/views.py
```diff
@@ -269,6 +264,17 @@ def change_password(request):
     return HttpResponse()
 
 
+@api_view(['GET'])
+def heartbeat(request):
```
shubham1g5, line 268: @snopoke What is the heartbeat endpoint mobile should be triggering based on this ?

snopoke: It will be `/users/heartbeat`

shubham1g5: And I guess `fcm_taken` is a query param here ? How do we communicate the Connect `user` info ?

snopoke: Correct, though this should probably be a POST request.

In terms of auth, this uses use token auth, same as 'change password'.

snopoke:
@sravfeyn FYI

utils/rest_framework.py
```diff
@@ -0,0 +1,31 @@
...
+        if not valid:
+            # Alternatively allow access tokens
+            # check if the request is valid and the protected resource may be accessed
+            valid, r = self.verify_request(request)
+            if valid:
+                request.resource_owner = r.user
+                return r.user, r.access_token
+            return None
+        else:
+            return OauthClientUser(), None
```
calellowitz, line 22: Are there any differences between this and the `django-oauth-toolkit` mixin in terms of validation logic, or just the method signature and return values?

utils/rest_framework.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ClientProtectedResourceAuth(OAuthLibMixin, BaseAuthentication):
+    """Authenticate request using Client credentials (as in the OAuth2 spec).
+    """
+
+    def authenticate(self, request):
+        # Validate either with HTTP basic or client creds in request body.
+        valid = self.authenticate_client(request)
+        if not valid:
```
calellowitz, line 13: Do we need this part if we want the view restricted to just client credentials? Allowing access tokens seems unnecessary here.

messaging/views.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Response:
+    {
+        "all_success": true,
+        "responses": [
+            {"status": "success", "username": "user1"},
+        ]
+    }
+    """
+    authentication_classes = [ClientProtectedResourceAuth]
```
calellowitz, line 28: If there are no scopes specified here, do no access tokens work or do any access tokens work?

calellowitz (commented):
Couple questions on the auth stuff, but otherwise looks great

utils/rest_framework.py
snopoke: No, the validation logic is the same. This just allows us to have DRF views with client credentials auth (thought I suppose we could still use the mixing and then configure the DRF view to allow unauthenticated requests.

utils/rest_framework.py
snopoke: True, I just copied this directly from the django-oauth-toolkit code.

messaging/views.py
snopoke: There are no permissions classes defined for this view and the default permission is `IsAuthenticated` so any authenticated request will pass.

calellowitz: Does that mean any user with a normal access token (the one they use for typical connectid operations or authentication with connect) can trigger these messages?

snopoke: Yes, I think that's true unless we remove the 'token' check

calellowitz: I think the only other view using this mixin specifies an "impossible" `required_scope` (one not in our settings file), so we would not be able to issue an access token with the relevant scope. I would be in favor of removing the token check, since I can't imagine we would need it, but adding a required scope here could also be fine.

utils/rest_framework.py
calellowitz: Based on the other conversation, I would be in favor of removing it, but will leave it up to you.

snopoke: Finally got round to this: 7f0b76d124ee5f0019d3dad9d88405d93939383a

calellowitz (approved):

PR #10, opened 2023-12-20
remove default validators

calellowitz:
This matches the live deployed code, but I apparently never made a PR for it.


pxwxnvermx (approved):
Thanks 🙌🏽

PR #11, opened 2024-01-10
Use django-environ to load Env variables

pxwxnvermx:
Added django-environ to load environment variables (db_url, secret etc.) from `.env` files.
`.env_example` file lists all the variables that need to be added.


requirements/requirements.txt
```diff
@@ -6,8 +6,6 @@
 #
 asgiref==3.6.0
     # via django
-build==0.10.0
```
calellowitz, line 9: do you know why there are so many requirements changes

calellowitz (commented):
As long as we know what is up with the requirements, this looks good

requirements/requirements.txt
pxwxnvermx: Yeah, pip-tools is not in the `requirements.in` file, so on pip-compile pip-tools and its dependencies were removed.

connectid/settings.py
```diff
@@ -196,10 +197,38 @@
...
+
+ALLOWED_HOSTS = ["127.0.0.1", "localhost"] + env.list(
+    "DJANGO_ALLOWED_HOSTS", default=[]
+)
+
+TWILIO_ACCOUNT_SID = env("TWILIO_ACCOUNT_SID")
+TWILIO_AUTH_TOKEN = env("TWILIO_AUTH_TOKEN")
+TWILIO_MESSAGING_SERVICE = env("TWILIO_MESSAGING_SERVICE")
+
+FCM_CREDENTIALS = env("FCM_CREDENTIALS", default=None)
```
sravfeyn, line 227: We should probably allow only individual fields within FCM_CREDENTIALs to be configured via env? (instead of the full JSON)

connectid/settings.py
```diff
@@ -196,10 +197,38 @@
     "DELETE_INACTIVE_DEVICES": False,
 }
 
-from .localsettings import *
```
sravfeyn, line 199: We are missing two settings from prod here `CSRF_TRUSTED_ORIGINS` and private-key part of  `OAUTH2_PROVIDER`

pxwxnvermx: Fixed. https://github.com/dimagi/connect-id/pull/11/commits/14d18c0a7ecb42c364fff716badee844679e1861

connectid/settings.py
pxwxnvermx: Fixed. https://github.com/dimagi/connect-id/pull/11/commits/14d18c0a7ecb42c364fff716badee844679e1861

pxwxnvermx:
These are included in https://github.com/dimagi/connect-id/pull/68

PR #16, opened 2024-02-16
Add otp generation and sending rate limiting

pxwxnvermx:
- Added a 2-minute time limit before sending the OTP SMS again
- New OTPs are generated only when the token is valid for 5 more minutes.


calellowitz (commented):
any chance you can add tests for this?

pxwxnvermx:
Yeah, i was working on adding tests for this. Will update this PR with the tests.

calellowitz (approved):

PR #17, opened 2024-02-22
Run connect-id using Docker

sravfeyn:


sravfeyn:
@calellowitz This is pretty much working and ready for review. If everything is good here, I can move to Ansible and then Kamal deploy

sravfeyn:
@calellowitz bump

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# install system dependencies
+RUN apt-get update && apt-get -y install libpq-dev gcc netcat
+
+# install dependencies
+RUN pip install --upgrade pip
+RUN /bin/bash
+
+# copy entrypoint.sh
+COPY ./entrypoint.sh /vendor/entrypoint.sh
+RUN sed -i 's/\r$//g' /vendor/entrypoint.sh
```
calellowitz, line 23: Why is this line necessary? Looks like it remove carriage returns, which I wouldn't think the file has

calellowitz (commented):
Other than the view changes, this looks good. Maybe a bad merge?

Dockerfile
sravfeyn: Yes, it can be removed. It's no longer necessary.

calellowitz:
Closing this since it was preempted by the linked PR above

PR #19, opened 2024-04-23
add recovery pin and urls

calellowitz:
This adds a recover PIN as an additional account recovery option

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -268,6 +268,36 @@ def change_password(request):
 
 
 @api_view(['POST'])
+def set_recovery_pin(request):
+    data = request.data
+    user = request.user
+    recovery_pin = data["recovery_pin"]
+    user.set_recovery_pin(recovery_pin)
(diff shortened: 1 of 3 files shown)


users/views.py
```diff
@@ -267,6 +267,36 @@ def change_password(request):
...
+    return HttpResponse()
+
+
+@api_view(['POST'])
+@permission_classes([])
+def confirm_recovery_pin(request):
+    data = request.data
+    phone_number = data["phone"]
+    secret_key = data["secret_key"]
+    user = ConnectUser.objects.get(phone_number=phone_number)
```
pxwxnvermx, line 286: Can we use `get_object_or_404` here? Right now this would cause a 500 due to ObjectNotFound error.

pxwxnvermx (approved):

calellowitz:
offlined with @pxwxnvermx, will fix in follow up

PR #23, opened 2024-05-03
Add Kamal deploy

sravfeyn:
https://dimagi.atlassian.net/browse/CCCT-317

I have deployed using this on a test server and it mostly works. 

The main difference in the way this works and commcare-connect deploy works is that this uses `kamal setup` to do initial setup while the commcare-connect uses ansible. 

Similarly for initial `.env` setup, commcare-connect uses ansible, here we have to setup the `.env` by hand. I will add automation for this in a latter PR.


sravfeyn:
@calellowitz bumping for review

config/deploy.yml
```diff
@@ -0,0 +1,110 @@
...
+
+# Name of the container image.
+image: connectid
+
+ssh:
+  user: ubuntu
+
+# Deploy to these servers.
+servers:
+  - 54.204.231.206
```
calellowitz, line 11: should this use our existing webworker?

config/deploy.yml
```diff
@@ -0,0 +1,110 @@
...
+  password: password: <%= %x(./registry_password.sh) %>
+
+
+builder:
+  multiarch: false
+  dockerfile: './Dockerfile'
+  context: './'
+
+healthcheck:
+  path: /admin/login/
```
calellowitz, line 25: Not entirely sure what the healthcheck does but `/users/` is effectively a healthcheck url that takes no auth and returns a 200 with a simple text response

connectid/settings.py
```diff
@@ -110,45 +110,34 @@
 
 AUTH_USER_MODEL = 'users.ConnectUser'
 
-LOG_DIR = BASE_DIR
-DJANGO_LOG_FILE = LOG_DIR / 'django.log'
 LOGGING = {
```
calellowitz, line 113: Can you explain a bit about the logging changes

calellowitz (commented):
Left a few questions but this looks great

config/deploy.yml
sravfeyn: Not right away. Once we are through review and 'migrate' the existing env to docker one.

config/deploy.yml
sravfeyn: It just checks whether the page loads. The default '/' one didn't work. But yeah, 'users/' might work

connectid/settings.py
sravfeyn: Sure. This mimics the logging strategy used by commcare-connect which uses the awslogs.  Basically instead of storing logs inside the docker containers that die, the logs go to console which then get sent to awslogs.

config/deploy.yml
```diff
@@ -0,0 +1,110 @@
...
+  port: 8000
+  max_attempts: 20
+  interval: 20s
+
+# Inject ENV variables into containers (secrets come from .env).
```
calellowitz, line 30: do we need all this commented out code?

calellowitz (approved):
Approving this with the understanding that it won't have any effect for now, since it is just merging into the other branch (and doesn't deploy to our live machines anyway). We can finally merge that one when we are confident in the end to end workflow

sravfeyn:
I am merging this into `sr/docker` https://github.com/dimagi/connect-id/pull/17/files

PR #27, opened 2024-05-14
add secondary recovery date and user key management

calellowitz:
This implements the web portion of [DB Key](https://docs.google.com/document/d/1VCFGdJ-2JvfSerMsg9QvKF3kCraqt_UrdE1DBu7S-Zw/edit) and [Secondary Phone Verification](https://docs.google.com/document/d/1CjpB6mSTF-tXJsdJIMmfUwCHmB7nWYFhZXtv0UHGWYw/edit)


users/models.py
```diff
@@ -36,6 +43,24 @@ def check_recovery_pin(self, pin):
...
 
+class UserKey(models.Model):
+    user = models.ForeignKey(ConnectUser, on_delete=models.CASCADE)
+    key = models.TextField()
+    valid = models.BooleanField(default=True)
+    created = models.DateTimeField(auto_now_add=True)
+
+    @classmethod
+    def get_or_create_key_for_user(cls, user):
+        now = now()
```
pxwxnvermx, line 54: Is this required? This is not getting used anywhere in the method.

users/models.py
```diff
@@ -36,6 +43,24 @@ def check_recovery_pin(self, pin):
         return check_password(pin, self.recovery_pin)
 
 
+class UserKey(models.Model):
+    user = models.ForeignKey(ConnectUser, on_delete=models.CASCADE)
+    key = models.TextField()
```
pxwxnvermx, line 48: Can we assign a max_length for this field? I ran the logic from `get_or_create_key_for_user` locally and after the `b64encode` the key length comes out to be 44 characters.

users/models.py
calellowitz: good catch

users/models.py
calellowitz: yeah, thats a good idea. will pad a bit for safety but a limit seems good

pxwxnvermx (approved):

PR #28, opened 2024-05-20
add basic credential models and views

calellowitz:
This is a first pass at the web part of credentials. I think it includes the core parts. I may include minor changes, but wanted to open it to get review started.


users/views.py
```diff
@@ -346,3 +346,56 @@ def get(self, request, *args, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        phone_numbers = request.data["users"]
+        org_name = request.data["organization_name"]
+        org_slug = request.data["organization"]
+        credential_name = request.data["credential"]
+        slug = f"{credential_name.lower().replace(' ', '_')}_{org_slug}"
+        credential, _ = Credential.objects.get_or_create(name=credential_name, organization_slug=org_slug, defaults={"credential_slug": slug})
+        users = ConnectUser.object.filter(phone_number__in=phone_numbers)
+        for u in users:
+            UserCredential.add_credential(user, credential, request)
```
sravfeyn, line 376: The underlying function seems to send an sms, is in intentional for this to be outside of an async background process?

users/models.py
```diff
@@ -63,3 +66,31 @@ class RecoverySteps(models.TextChoices):
...
+    user = models.ForeignKey(ConnectUser, on_delete=models.CASCADE)
+    credential = models.ForeignKey(Credential, on_delete=models.CASCADE)
+    accepted = models.BooleanField(default=False)
+    invite_id = models.CharField(max_length=50, default=uuid4, unique=True)
+
+    class Meta:
+        unique_together = ("user", "credential")
+
+    @classmethod
+    def add_credential(cls, user, credential, request):
```
sravfeyn, line 87: I can't think of better names for this method either, but I wouldn't be able guess that this would send an SMS to the user from the method name

users/views.py
```diff
@@ -346,3 +346,56 @@ def get(self, request, *args, **kwargs):
         demo_users = PhoneDevice.objects.filter(phone_number__startswith=TEST_NUMBER_PREFIX).values('phone_number', 'token')
         results = {"demo_users": list(demo_users)}
         return JsonResponse(results)
+
+
+class FilterUsers(ClientProtectedResourceMixin, View):
+    required_scopes = ['user_fetch']
+
+    def get(self, request, *args, **kwargs):
+        country = request.data["country"]
```
pxwxnvermx, line 355: This should be `request.GET` instead of `request.data`.

users/views.py
```diff
@@ -346,3 +346,56 @@ def get(self, request, *args, **kwargs):
...
+    )
+
+
+
+class FetchCredentials(ClientProtectedResourceMixin, View):
+    required_scopes = ['user_fetch']
+
+    def get(self, request, *args, **kwargs):
+        credentials = Credential.objects.all().values('name', 'slug')
+        return JsonResponse(list(credentials))
```
pxwxnvermx, line 401: same as above.

users/models.py
calellowitz: That is fair, I will think of better names, but will leave this for now as QA gets underway

users/views.py
calellowitz: I changed this to return a dict, rather than a list. You will need to update the connect part to accomodate that but it seemed better than passing an unsafe flag.

users/views.py
calellowitz: That is a good point. I had assumed it was hit from a celery task in connect. That turns out not to be true, but I think it makes the most sense to hit it from one there, and have the return value here confirm that all were sent. Otherwise, the caller can't be sure the credentials were added properly.

calellowitz: fyi @pxwxnvermx

users/views.py
```diff
@@ -339,10 +340,64 @@ def get(self, request, *args, **kwargs):
...
+    )
+
+
+
+class FetchCredentials(ClientProtectedResourceMixin, View):
+    required_scopes = ['user_fetch']
+
+    def get(self, request, *args, **kwargs):
+        credentials = Credential.objects.all().values('name', 'slug')
+        results = {"credentials":  list(crendentials)}
```
pxwxnvermx, line 402: There is a typo here `crendentials` -> `credentials`.

users/views.py
```diff
@@ -339,10 +340,64 @@ def get(self, request, *args, **kwargs):
...
+class GetDemoUsers(ClientProtectedResourceMixin, APIView):
     required_scopes = ['user_fetch']
 
     def get(self, request, *args, **kwargs):
         demo_users = PhoneDevice.objects.filter(phone_number__startswith=TEST_NUMBER_PREFIX).values('phone_number', 'token')
         results = {"demo_users": list(demo_users)}
         return JsonResponse(results)
+
+
+class FilterUsers(ClientProtectedResourceMixin, APIView):
```
pxwxnvermx, line 352: The ClientProtectedResourceMixin does not work with APIView, which leads to the requests for this view to be unauthorized.

users/views.py
```diff
@@ -339,10 +340,64 @@ def get(self, request, *args, **kwargs):
...
+        country = request.query_params.get("country")
+        credential = request.query_params.get("credential")
+        users = UserCredential.objects.filter(credential__slug=credential, user__phone_number__startswith=country, accepted=True).select_related('user')
+        user_list = [{"username": u.user.username, "phone_number": u.user.phone_number, "name": u.user.name} for u in users]
+        result = {"found_users": user_list}
+        return JsonResponse(result)
+        
+
+
+class AddCredential(ClientProtectedResourceMixin, APIView):
```
pxwxnvermx, line 365: same as above.

users/views.py
```diff
@@ -339,10 +340,64 @@ def get(self, request, *args, **kwargs):
...
+    required_scopes = ['user_fetch']
+
+    def post(self, request, *args, **kwargs):
+        phone_numbers = request.data["users"]
+        org_name = request.data["organization_name"]
+        org_slug = request.data["organization"]
+        credential_name = request.data["credential"]
+        slug = f"{credential_name.lower().replace(' ', '_')}_{org_slug}"
+        credential, _ = Credential.objects.get_or_create(name=credential_name, organization_slug=org_slug, defaults={"slug": slug})
+        users = ConnectUser.object.filter(phone_number__in=phone_numbers)
```
pxwxnvermx, line 375: typo
```suggestion
        users = ConnectUser.objects.filter(phone_number__in=phone_numbers)
```

OrangeAndGreen (approved):

PR #35, opened 2024-06-12
Fix filter_users conditions should be OR'd

pxwxnvermx:
The current country and credential filters on the filter_users API need both country and credential filters to be not empty to return data. This PR fixes them and makes them work independently as well.


calellowitz:
I don't think we want to `OR` (we dont want all users in a country OR credentialed users if both are given), but I fixed it to properly handle the situation if only one is given in https://github.com/dimagi/connect-id/pull/36

PR #38, opened 2024-07-08
User Deactivation Views

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-374)


calellowitz (commented):
couple small things. The big one I made on the other PR but i think that connectid should call the connect view, not the mobile

users/oauth.py
```diff
@@ -1,10 +1,17 @@
 from oauth2_provider.oauth2_validators import OAuth2Validator
 
+
 class ConnectOAuth2Validator(OAuth2Validator):
+    oidc_claim_scope = OAuth2Validator.oidc_claim_scope
```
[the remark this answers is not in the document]
pxwxnvermx: This copies the claim scopes from the base class to our oauth2 validator class. Before only the `sub` was transferred when the userinfo endpoint was called. This allows us to get additional information on the claim.

hemant10yadav (approved):

users/views.py
```diff
@@ -423,3 +431,48 @@ def get(self, request, *args, **kwargs):
...
         results = {"credentials":  list(credentials)}
         return JsonResponse(results)
+
+
+@api_view(["POST"])
+@permission_classes([])
+def initiate_deactivation(request):
+    data = request.data
+    phone_number = data["phone_number"]
+    secret_key = data["secret_key"]
```
calellowitz, line 441: when does this secret key get set?

users/views.py
```diff
@@ -423,3 +431,48 @@ def get(self, request, *args, **kwargs):
...
+    try:
+        phone_device = PhoneDevice(phone_number=phone_number)
```
calellowitz, line 443: it looks like you are intending to fetch from the data base, but instead creating a new object. Also, if you did intend to query, this will also return devices if this number was ever used as a secondary number (for example if one worker put another's as their recovery number). I think you want to be fetching the user directly by phone number, like we do in the other recovery views, rather than the `PhoneDevice`

users/views.py
```diff
@@ -423,3 +431,48 @@ def get(self, request, *args, **kwargs):
...
+
+@api_view(["POST"])
+@permission_classes([])
+def confirm_deactivation(request):
+    data = request.data
+    phone_number = data["phone_number"]
+    secret_key = data["secret_key"]
+    deactivation_token = data["token"]
+    try:
+        phone_device = PhoneDevice(phone_number=phone_number)
```
calellowitz, line 462: same comment about this as above

users/models.py
```diff
@@ -9,14 +10,16 @@
...
 from phonenumber_field.modelfields import PhoneNumberField
 
 from utils import get_sms_sender, send_sms
 
 from .const import TEST_NUMBER_PREFIX
-# Create your models here.
+
 
 class ConnectUser(AbstractUser):
     phone_number = PhoneNumberField(unique=True)
```
calellowitz, line 25: sravan mentioned this on one of the specs, but we will need to update this constraint so that phone number is unique for active users, but not overall. i think you can do that by adding a unique constraint with `condition` https://docs.djangoproject.com/en/5.0/ref/models/constraints/#condition

calellowitz (changes requested):
general approach looks good but i think there are a few bugs.

users/models.py
pxwxnvermx: Added. https://github.com/dimagi/connect-id/pull/38/commits/7a81ac5dcd6eb7471a97a3168321f932be073e45

users/views.py
pxwxnvermx: Fixed. Thanks. I did intend to query that. https://github.com/dimagi/connect-id/pull/38/commits/9bce35ba1fac9c8a08a079d8e1a6ad51514d8ecf

users/views.py
pxwxnvermx: This secret_key is the one from recovery status.

calellowitz (approved):

PR #41, opened 2024-09-05
Added models for messaging

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-474


calellowitz (commented):
It might be easier to just combine this with the other PR, which would make review easier, and if changes affect the models, you wont have to update two PRs at once.

connectmessaging/models.py
```diff
@@ -0,0 +1,24 @@
...
+    connect_user = models.ForeignKey(ConnectUser, on_delete=models.CASCADE)
+    key_url = models.URLField(max_length=200)
+    callback_url = models.URLField(max_length=200)
+    delivery_url = models.URLField(max_length=200)
+
+
+class Message(models.Model):
+    message_id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    channel = models.ForeignKey(Channel, on_delete=models.CASCADE)
+    content = models.BinaryField()
```
calellowitz, line 22: This can be a text field (the message will be base64 encoded)

hemant10yadav: Ok I will discard  and raise a combine PR.

PR #42, opened 2024-09-05
Stub out ConnectID views

hemant10yadav:
[CCCT-474](https://dimagi.atlassian.net/browse/CCCT-474)
[CCCT-473](https://dimagi.atlassian.net/browse/CCCT-473)

[CCCT-474]: https://dimagi.atlassian.net/browse/CCCT-474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-473]: https://dimagi.atlassian.net/browse/CCCT-473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


connectmessaging/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class ConnectmessagingConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "connectmessaging"
```
sravfeyn, line 6: Why not use the existing `messaging` app?

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+
+from django.db.models import Prefetch
+from django.shortcuts import get_object_or_404
+from django.utils import timezone
+from rest_framework import status, views
+from rest_framework.authentication import BasicAuthentication
+from rest_framework.exceptions import ValidationError
+from rest_framework.response import Response
+
+from messaging.serializers import Message as Msg
```
sravfeyn, line 11: `nit` Recommend to use just Message instead of Msg

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+    authentication_classes = [ClientProtectedResourceAuth]
+
+    def post(self, request, *args, **kwargs):
+        serializer = ChannelSerializer(data=request.data)
+        if serializer.is_valid():
+            channel = serializer.save()
+            message = Msg(
+                usernames=[channel.connect_user.username],
+                title="Channel created",
+                body="A new channel has been created for you. Please provide your consent to proceed.",
```
sravfeyn, line 34: If this is a user-facing message, we should consider reframing this to on the lines of 'Please provide consent to send/receive messages', users don't need to be aware of the `channel` term

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+                data={"keyUrl": channel.key_url},
+            )
+            send_bulk_message(message)
```
sravfeyn, line 37: Should this handle message failure?

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+                )
+
+                if is_request_from_hq(request):
+                    send_bulk_message(message_to_send)
+                else:
+                    if not Channel.objects.filter(
+                            channel_id=channel.channel_id, user_consent=True
+                    ).exists():
+                        return Response(
+                            {"error": "Consent is required for this channel."},
```
sravfeyn, line 76: It might make sense to send some sort of codenames for the caller to identify the error types.

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+                    data={"message_id": message.message_id, "content": message.content},
+                )
+
+                if is_request_from_hq(request):
+                    send_bulk_message(message_to_send)
+                else:
+                    if not Channel.objects.filter(
+                            channel_id=channel.channel_id, user_consent=True
+                    ).exists():
+                        return Response(
```
sravfeyn, line 75: Should this be a `JsonResponse`?

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+                channel = message.channel
+                message_to_send = Msg(
+                    usernames=[channel.connect_user.username],
+                    data={"message_id": message.message_id, "content": message.content},
+                )
+
+                if is_request_from_hq(request):
+                    send_bulk_message(message_to_send)
+                else:
+                    if not Channel.objects.filter(
```
sravfeyn, line 72: It might be worth refactoring the above `for` loop to avoid separate query for each message channel by making a single query to see if there are any missing channels.

connectmessaging/views.py
sravfeyn, line 76: This will result in partial success/failure. For e.g. if there are five messages to be sent and 3rd channel doesn't exist, first two messages will go and last three messages won't go even if the last two messages have channel. 
I am not sure what's required, whether we should fail on all messages or allow partial success. And if we allow partial success, whether that needs to be informed to caller.

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+        channels_data = ChannelSerializer(channels, many=True).data
+
+        messages = []
+        for channel in channels:
+            channel_messages = channel.message_set.all()
+            messages.extend(channel_messages)
+
+        messages_data = MessageSerializer(messages, many=True).data
+
+        return Response({"channels": channels_data, "messages": messages_data})
```
sravfeyn, line 111: Shouldn't this be paginated?

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+class UpdateConsentView(views.APIView):
```
sravfeyn, line 114: How about permissions/auth? What permissions are needed to make this call?

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+
+        #: TO-DO update the url
+        url = "CONSENT_URL"
+        json_data = {
+            "channel_id": str(channel.channel_id),
+            "consent": str(channel.user_consent),
+        }
+        make_request_to_service(url=url, json_data=json_data)
+
+        return Response(status=status.HTTP_200_OK)
```
sravfeyn, line 135: Should be a `JsonResponse`

connectmessaging/views.py
```diff
@@ -0,0 +1,184 @@
...
+        message_ids = request.data.get("messages", [])
+
+        messages = (
+            Message.objects.select_for_update()
+            .filter(message_id__in=message_ids)
+            .select_related("channel")
+        )
+
+        if not messages.exists():
+            return Response(status=status.HTTP_200_OK)
```
sravfeyn, line 149: Should this be a `404`?

connectmessaging/views.py
sravfeyn: Should probably send a status URL which the client can check for the status. @calellowitz ?

connectmessaging/views.py
hemant10yadav: To avoid confusion with another Message model in the connectmessaging app, I made the following adjustments.

connectmessaging/views.py
hemant10yadav: You are right, but as I understand it, this endpoint will receive a single message from a mobile device, so it will handle just one message in that case. However, it could receive multiple messages when this endpoint is used by HQ. That's why I made it a list to create a common view, but I am not fully confident that only one message will be sent from a mobile. Perhaps @calellowitz can clarify this.

connectmessaging/views.py
hemant10yadav: [same](https://github.com/dimagi/connect-id/pull/42/files/aceba49c0a87618de4404b516d4a807a6fe42abe#r1802637970)

connectmessaging/views.py
hemant10yadav: If no `authentication_classes` are provided, the default authentication classes will be used.

Verified that APIView automatically uses DEFAULT_AUTHENTICATION_CLASSES
```
class APIView(View):
    authentication_classes = api_settings.DEFAULT_AUTHENTICATION_CLASSES
```

connectmessaging/views.py
hemant10yadav: Added the change.

connectmessaging/views.py
hemant10yadav: Thanks Changed to `JsonResponse`.

connectmessaging/apps.py
calellowitz: As mentioned in our call, I agree with this

connectmessaging/views.py
calellowitz: I think you interpreted that backwards. Mobile will send many at a time (whenever it gets access to network), while HQ will send them individually since it can send them as soon as they are needed.

connectmessaging/views.py
calellowitz: Since this is only supposed to send undelivered messages, and there aren't likely to be too many of those, I think it's fine to have it unpaginated for now. If we run into issues with a huge message backlog, we can reconsider.

connectmessaging/views.py
calellowitz: Check which status?

connectmessaging/views.py
```diff
@@ -0,0 +1,193 @@
...
+def is_request_from_hq(request):
+    host = request.META.get("HTTP_HOST")
+    return host == "commcarehq.org"
+
+
+class CreateChannelView(views.APIView):
+    authentication_classes = [ClientProtectedResourceAuth]
+
+    def post(self, request, *args, **kwargs):
+        serializer = ChannelSerializer(data=request.data)
```
calellowitz, line 28: Is there an advantage to doing this over just using the model directly `Channel(...`?

connectmessaging/views.py
calellowitz: If this fails, it should probably bubble back to hq, so it knows it needs to create it again.

connectmessaging/views.py
```diff
@@ -0,0 +1,193 @@
...
+    ]
+
+    def post(self, request, *args, **kwargs):
+        data = None
+        if isinstance(request.data, list):
+            data = request.data
+        else:
+            data = [request.data]
+
+        serializer = MessageSerializer(data=data, many=True)
```
calellowitz, line 59: Same question about using the serializer here vs the model directly

connectmessaging/views.py
```diff
@@ -0,0 +1,193 @@
...
+        if serializer.is_valid():
+            messages = serializer.save()
+            for message in messages:
+                channel = message.channel
+                message_to_send = Msg(
```
calellowitz, line 64: Does this class work for messages to HQ and to FCM? I would have imagined that an FCM specific class like that would have a different data model than what we send to hq

connectmessaging/views.py
```diff
@@ -0,0 +1,193 @@
...
+                    usernames=[channel.connect_user.username],
+                    data={"message_id": message.message_id, "content": message.content},
+                )
+
+                if is_request_from_hq(request):
```
calellowitz, line 69: This isn't safe since we will have other sending services like OCS. The last time we chatted we discussed having different URLs from mobile and from sending services. I think that is probably the best idea since they will likely need different data and actions anyway.

connectmessaging/task.py
```diff
@@ -0,0 +1,21 @@
+import requests
+
+from connectid import celery_app
+
+
+class CommCareHQAPIException(Exception):
+    pass
+
+
+@celery_app.task()
```
calellowitz, line 10: Let's leave this synchronous for now without celery for our testing. And we can add async before going live

calellowitz (changes requested):
Left a few comments, both to agree with previous reviews and a few more suggestions. In addition, it would be good to add the `Server` model we discussed on the call today

connectmessaging/views.py
hemant10yadav: My bad I will update it accordingly.

connectmessaging/views.py
hemant10yadav: We need to extract keys from the data and catch any channel exceptions, so I thought using a serializer would result in much cleaner and more maintainable code than using the model directly. However, I can switch to using Channel(...) if you think that's the better approach.

connectmessaging/views.py
hemant10yadav: This class only work for FCM notification. I think I accidentally placed it outside .

calellowitz (commented):
Generally looks fine, left a couple comments about naming and one about the behavior of the message

messaging/views.py
```diff
@@ -139,3 +158,189 @@ def _build_notification(data):
...
+            )
+            send_bulk_message(message_to_send)
+            return JsonResponse(
+                {"message_id": str(message.message_id)},
+                status=status.HTTP_200_OK,
+            )
+        return JsonResponse(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class SendMessageView(APIView):
```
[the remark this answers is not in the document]
hemant10yadav: Thanks! updated [3a6dfcb](https://github.com/dimagi/connect-id/pull/42/commits/3a6dfcbb47257807cf645a01b798b7885403a259)

messaging/views.py
```diff
@@ -139,3 +158,189 @@ def _build_notification(data):
...
+
+    def post(self, request, *args, **kwargs):
+        data = request.data
+        serializer = MessageSerializer(data=data)
+        if serializer.is_valid():
+            message = serializer.save()
+            channel = message.channel
+            message_to_send = MessageData(
+                usernames=[channel.connect_user.username],
+                data={"message_id": message.message_id, "content": message.content},
```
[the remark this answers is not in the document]
hemant10yadav: Added [3a6dfcb](https://github.com/dimagi/connect-id/pull/42/commits/3a6dfcbb47257807cf645a01b798b7885403a259)

messaging/views.py
```diff
@@ -139,3 +158,189 @@ def _build_notification(data):
...
+            data={"keyUrl": server.key_url},
+        )
+        # send fcm notification.
+        send_bulk_message(message)
+        return JsonResponse(
+            {"channel_id": str(channel.channel_id)}, status=status.HTTP_201_CREATED
+        )
+
+
+class SendFcmNotificationView(APIView):
```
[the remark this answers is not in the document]
hemant10yadav: Chaged [3a6dfcb](https://github.com/dimagi/connect-id/pull/42/commits/3a6dfcbb47257807cf645a01b798b7885403a259)

messaging/tests.py
```diff
@@ -110,7 +124,377 @@ def test_send_message_bulk(authed_client, fcm_device):
...
+                "messaging.views.send_messages_to_service_and_mark_status"
+        ) as mock_send_bulk_message:
+            response = auth_device.post(
+                self.url,
+                data=json.dumps(data),
+                content_type="application/json",
+            )
+            json_data = response.json()
+
+            assert (
```
sravfeyn, line 281: We don't need the assertion fail-part of these statements. If the tests fail, it will pretty much say the same along the lines of these statements.

messaging/tests.py
```diff
@@ -110,7 +124,377 @@ def test_send_message_bulk(authed_client, fcm_device):
...
+                    response.status_code == status.HTTP_201_CREATED
+            ), f"Expected status {status.HTTP_201_CREATED}, but got {response.status_code}."
+            assert (
+                    "message_id" in json_data
+            ), "Expected 'message_id' in response data, but it was missing."
+
+            message_ids = json_data["message_id"]
+            assert (
```
sravfeyn, line 289: We can skip the two asserts above since we have this here, right?

messaging/tests.py
```diff
@@ -110,7 +124,377 @@ def test_send_message_bulk(authed_client, fcm_device):
...
+        response = auth_device.get(self.url)
+        json_data = response.json()
+
+        assert (
+                response.status_code == status.HTTP_200_OK
+        ), f"Expected status code 200, but got {response.status_code}"
+        assert "channels" in json_data, "Response is missing 'channels' key"
+        assert "messages" in json_data, "Response is missing 'messages' key"
+        assert len(json_data["channels"]) == 0, "Expected empty channels list"
+        assert len(json_data["messages"]) == 0, "Expected empty messages list"
```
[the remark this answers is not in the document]
hemant10yadav: Thanks! added the changes.

calellowitz (approved):
approving and merging for now so we can start testing, as issues come up we may need to make adjustments

PR #48, opened 2024-11-06
messaging auth

calellowitz:
There were two issues in the plan to use oauth apps for auth on messaging and request signing that recently surfaced.

1) The client secret was stored hashed like a password so could not be used for request signing because connectid didn't have access to the key value (it was no longer a shared secret).
2) There is no straightforward way with our oauth libraries to restrict client credential grant applications to specific scopes which meant that messaging servers would have had access to the rest of the OAuth protected views, including the user fetch and credential views.

This changes to an effectively equivalent auth (client ids and secrets generated in the same way as our oauth lib), but stored on the messaging server model and restricted to only messaging.


utils/rest_framework.py
```diff
@@ -21,3 +21,23 @@ def is_authenticated(self):
...
     def __str__(self):
         return "OauthClientUser"
+
+
+class MessagingServerAuth(OAuthLibMixin, BasicAuthentication):
+    """Authenticate request using Client credentials (as in the OAuth2 spec).
+    """
+
+    def authenticate_credentials(self, userid, password, request=None):
+        server = MessageServer.objects.get(server_id=userid)
```
snopoke, line 31: needs error handling

messaging/models.py
```diff
@@ -3,17 +3,22 @@
...
 
 class MessageServer(models.Model):
     name = models.CharField(max_length=255)
-    oauth_application = models.ForeignKey(Application, on_delete=models.CASCADE)
     key_url = models.URLField(max_length=200)
     callback_url = models.URLField(max_length=200)
     delivery_url = models.URLField(max_length=200)
     consent_url = models.URLField(max_length=200)
+    server_id = models.CharField(max_length=100, unique=True, default=generate_client_id, db_index=True)
+    secret_key = models.CharField(
```
snopoke, line 18: Have you considered encrypting this since this isn't hashed any longer?

On OCS we use https://pypi.org/project/django-cryptography-django5/

users/views.py
```diff
@@ -104,7 +104,7 @@ def confirm_secondary_otp(request):
     # check otp code for user
     # mark phone as confirmed on user model
     user = request.user
-    device = PhoneDevice.objects.get(phone_number=user.recovery_phone, user=user)
+    device, _ = PhoneDevice.objects.get(phone_number=user.recovery_phone, user=user)
```
pxwxnvermx, line 107: The `_` might cause an error here as `.get` only returns a single object. Maybe you meant to use `.get_or_create` here?

calellowitz: yep, that is definitely what I intended

messaging/models.py
calellowitz: I hadn't but that seems like a reasonable idea. Given it's unencrypted storage elsewhere (in hq settings), I had imagined this was sufficient for now, but I will look into this library, seems easy enough.

utils/rest_framework.py
calellowitz: added

pxwxnvermx (approved):

calellowitz:
I can follow up with the encryption, merging for testing

PR #51, opened 2024-11-11
endpoints to receive/forward hq invites

sravfeyn:
[Spec](https://docs.google.com/document/d/1h9OlQDD7Jpygwyi-Zp6aU-_TQNkz4wOxw93ZOjkTwHY/edit?tab=t.0#heading=h.syj1ugk3stxe)
[Ticket](https://dimagi.atlassian.net/browse/CCCT-425)


users/views.py
```diff
@@ -417,6 +419,52 @@ def post(self, request, *args, **kwargs):
...
+        sender = get_sms_sender(user.phone_number.country_code)
+        send_sms(user.phone_number.as_e164, message, sender)
+        return JsonResponse({"success": True})
+
+
+class ConfirmHQInviteCallback(APIView):
+    authentication_classes = [ClientProtectedResourceAuth]
+
+    def post(self, request, *args, **kwargs):
+        callback_url = quote_plus(request.data["callback_url"])
```
calellowitz, line 456: where does `callback_url` come from and point to?

users/views.py
```diff
@@ -417,6 +419,52 @@ def post(self, request, *args, **kwargs):
...
+        Thanks.
+        -The ConnectID Team.
+        """
+        sender = get_sms_sender(user.phone_number.country_code)
+        send_sms(user.phone_number.as_e164, message, sender)
+        return JsonResponse({"success": True})
+
+
+class ConfirmHQInviteCallback(APIView):
+    authentication_classes = [ClientProtectedResourceAuth]
```
calellowitz, line 453: Who makes this call? Is this not from mobile?

calellowitz (commented):
Couple small questions

users/views.py
sravfeyn: Yes, this is made by mobile.

# From spec
To accept invitation (when user confirms), call below endpoint

‘connectid.dimagi.com/users/confirm_hq_invite’

POST params
callback_url: callback_url (forwarded in deep link)
invite_code: invite_code (forwarded in deep link)
user_token: user’s active auth token

users/views.py
sravfeyn: This is all according to the modified spec.

This comes from HQ and it points to HQ's invite confirmation view. It's for actual invite confirmation, when mobile confirms it to ConnectID, ConnectID calls back HQ on callback_url with the user's token.

sravfeyn: This is the HQ change https://github.com/dimagi/commcare-hq/pull/34967/commits/407303f6c68034fd182f7325a2764eb3f608e74b

sravfeyn:
@calellowitz Please let me know if this is good to go.

users/views.py
```diff
@@ -417,6 +419,52 @@
...
+class ProcessHQInviteCallback(APIView):
+    authentication_classes = [ClientProtectedResourceAuth]
+
+    def post(self, request, *args, **kwargs):
+        callback_url = quote_plus(request.data["callback_url"])
+        invite_code = request.data["invite_code"]
+        user_token = request.data["user_token"]
+        invite_code = unquote_plus(callback_url)
+        try:
+            response = requests.post(callback_url, data={"invite_code": invite_code, "token": user_token})
```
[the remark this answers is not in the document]
calellowitz: Are you planning to address this? Given your answer below about where this points, it should be doable to check it against an allowed list (only hq instances?)

users/views.py
calellowitz: In that case, this auth is incorrect. It only works with oauth client credentials grants (what we use for connect or other servers). I think the default auth classes should be fine.

connectid/views.py
```diff
@@ -28,3 +28,10 @@ def assetlinks_json(request):
         },
     ]
     return JsonResponse(assetfile, safe=False)
+
+
+def deeplink(request, subpath):
```
calellowitz, line 33: Are we able to direct them directly to the play store? That is what we discussed on the call and seems to make the most sense if it is possible. Mobile seemed to believe that it would be.

users/views.py
```diff
@@ -434,7 +434,8 @@ def post(self, request, *args, **kwargs):
             # We don't want to make this a user lookup service
             # So fake a success message
             return JsonResponse({"success": True})
-        deeplink = f"https://connectid.dimagi.com/hq_invite/{callback_url}/{username}/{invite_code}/{user_domain}"
+        details = f"{callback_url}/{username}/{invite_code}/{user_domain}/{user.username}"
```
calellowitz, line 437: I think we spoke about this a while ago, but why is this encoded as a path rather than using query params (which libraries on web and mobile could parse much more easily)?

calellowitz (changes requested):
Left a few questions on the new code in addition to the previous comments

connectid/views.py
sravfeyn: Right, I have updated it to redirect to the app.

Screenshot 2024-11-22 at 12 31 38 PM

users/views.py
sravfeyn: Yeah, I think there was some issue regarding formatting this using query params on mobile, this format was specified by mobile.

users/views.py
sravfeyn: You are right! I will remove those classes.

sravfeyn:
@calellowitz Thanks for the review, I have followed up.

connectid/views.py
```diff
@@ -31,7 +31,42 @@ def assetlinks_json(request):
 
 
 def deeplink(request, subpath):
-    # If mobile is not able to process the deeplink
-    #   and the URL is opened in the browser, ask the
-    #   user to install the app and try again
-    return HttpResponse("Please install CommCare and re-open this link to accept invite")
+    html_content = """
```
calellowitz, line 34: Why inline HTML instead of using a separate file

connectid/settings.py
```diff
@@ -80,7 +80,10 @@
 WSGI_APPLICATION = 'connectid.wsgi.application'
 
 
-
+TRUSTED_COMMCAREHQ_HOSTS = [
+    "commcarehq.org",
```
calellowitz, line 84: how does this work? Specifically, if `commcarehq.org` is enough to catch `www.commcarehq.org` I would expect it to also catch staging, and if we need to be more specific, I would expect `www` to be required.

users/views.py
calellowitz: I spoke to @OrangeAndGreen and he seemed to believe query params would be better for them as well. Let's offline on slack about this with a mobile dev.

connectid/views.py
sravfeyn: I have moved it to its own dir.

connectid/settings.py
sravfeyn: Good point, I will add `www` to the list to be specific.

users/views.py
sravfeyn: Okay, I have updated this to use queryparams.

sravfeyn:
@calellowitz I have addressed your feedback, please have a look. I think the CodeQL failure is incorrect as it is not able to detect the trusted URL check that is added.

sravfeyn:
@calellowitz Bumping for review please.

calellowitz (approved):

sravfeyn:
Thanks @calellowitz please also review the corresponding HQ PR https://github.com/dimagi/commcare-hq/pull/34967/files#diff-dc2c79b556e950acc7dea90175fd5febba22bf9fa4faaba1b61c3b3d5acdc9e9

PR #68, opened 2025-01-08
CI Workflow

hemant10yadav:
This PR introduces the following changes: [CCCT-637](https://dimagi.atlassian.net/browse/CCCT-637)

- Adds **django-environ** to load environment variables.
- Includes **ci.yml** configuration for running **pytest** in CI.
- Adds **linting** configurations.

**Additionally, follow the branches below to view formatting and pytest fixes:**

- Formatting changes: [#69](https://github.com/dimagi/connect-id/pull/69)
- Test fixes: [#70](https://github.com/dimagi/connect-id/pull/70)

[CCCT-637]: https://dimagi.atlassian.net/browse/CCCT-637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


connectid/settings.py
```diff
@@ -267,3 +242,62 @@
...
+        "openid": "OpenID Connect scope",
+        "sync": "sync with commcarehq"
+    },
+    "PKCE_REQUIRED": False,
+    "OAUTH2_VALIDATOR_CLASS": "users.oauth.ConnectOAuth2Validator",
+}
+
+fcm_private_key = env("FCM_PRIVATE_KEY", default="")
+
+if fcm_private_key:
```
calellowitz, line 284: nit: should be all caps

connectid/settings.py
```diff
@@ -267,3 +242,62 @@
...
+    FCM_CREDENTIALS = {
+        "type": "service_account",
+        "project_id": env("FCM_PROJECT_ID", default=""),
+        "private_key_id": env("FCM_PRIVATE_KEY_ID", default=""),
+        "private_key": env("FCM_PRIVATE_KEY", default=""),
```
calellowitz, line 289: nit: this is alread set above

connectid/settings.py
```diff
@@ -235,19 +219,10 @@
 SENTRY_ENVIRONMENT = "local"
```
calellowitz, line 219: these sentry variables (and app hash above) need to be pulled from the env too

connectid/settings.py
```diff
@@ -267,3 +242,62 @@
...
+
+SECRET_KEY = env(
+    "SECRET_KEY",
+    default="django-insecure-yofpqrszrdtv0ftihjd09cuim2al9^n9j^b85%-y0v*^_lj18d",
+)
+
+# SECURITY WARNING: don't run with debug turned on in production!!
+DEBUG = env("DEBUG", default=False)
+
+DATABASES = {
```
calellowitz, line 254: This needs to be set with env variables to be able to connect to the DB

calellowitz (changes requested):
Added a few changes but this looks very close

calellowitz (commented):
Looks like a few changes still missing, but getting closer

connectid/settings.py
hemant10yadav: We are loading `DATABASE_URL` from env right?

calellowitz: Ah I see, they want a URL with all the username password info. Do you know what the advantage of this over manually constructing a dict is? It seems fine, but I am not sure I understand the pros and cons

.env_template
```diff
@@ -19,7 +19,7 @@ FCM_PRIVATE_KEY=
 
 OIDC_RSA_PRIVATE_KEY=
 
-APP_HASH="apphash"
+APP_HASH=""
```
calellowitz, line 22: nit: why empty string, not none?

connectid/settings.py
```diff
@@ -213,11 +213,11 @@
 
 SITE_ID = 1
 
-APP_HASH = env("APP_HASH")
+APP_HASH = env("APP_HASH", default="apphash")
 
-SENTRY_DSN = env("SENTRY_DSN", default=None)
+SENTRY_DSN = env("SENTRY_DSN", default="")
```
calellowitz, line 218: why the type change?

pyproject.toml
```diff
@@ -5,3 +5,50 @@ python_files = [
...
+[tool.black]
+line-length = 119
+target-version = ['py311']
+
+
+# ==== isort ====
+[tool.isort]
+profile = "black"
+line_length = 119
+known_first_party = [
```
pxwxnvermx, line 19: This key needs to be removed from this config. It is used to force "isort" to recognize a folder as a module in the current project. Since we have multiple app modules in the base folder, we would also need to update this every time we create a new app. The formatting and lining should work without this key as well.

https://pycqa.github.io/isort/docs/configuration/options.html#known-first-party

pyproject.toml
```diff
@@ -5,3 +5,50 @@ python_files = [
...
+skip = ["venv/"]
+skip_glob = ["**/migrations/*.py"]
+
+# ==== PyLint ====
+[tool.pylint.MASTER]
+load-plugins = [
+    "pylint_django",
+    "pylint_celery",
+]
+django-settings-module = "config.settings.local"
```
pxwxnvermx, line 32: This also requires `__init__.py` file in `connectid` folder to make it a python module.

```suggestion
django-settings-module = "connectid.settings"
```

pyproject.toml
hemant10yadav: Thanks will remove it.

.env_template
hemant10yadav: Thanks I was trying something then forgot to remove it.

connectid/settings.py
```diff
@@ -267,3 +242,62 @@
...
+
+CSRF_TRUSTED_ORIGINS = env.list("CSRF_TRUSTED_ORIGINS", default=[])
+
+TWILIO_ACCOUNT_SID = env("TWILIO_ACCOUNT_SID", default=None)
+TWILIO_AUTH_TOKEN = env("TWILIO_AUTH_TOKEN", default=None)
+TWILIO_MESSAGING_SERVICE = env("TWILIO_MESSAGING_SERVICE", default=None)
+
+OAUTH2_PROVIDER = {
+    "OIDC_ENABLED": True,
+    "OIDC_RSA_PRIVATE_KEY": env("OIDC_RSA_PRIVATE_KEY", default=""),
```
sravfeyn, line 273: Have you tested that OAuth works okay? I remember having to surround this inside quotes [here](https://github.com/dimagi/connect-id/pull/17/files#diff-d144ad634a13af213dd8b4d719b4debda4bfc4b1525a9456183ea13a84adb416R171)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,63 @@
...
+  pytest:
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:12
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_PASSWORD: postgres
```
sravfeyn, line 42: You don't need `POSTGRES_USER`?

sravfeyn:
Apart from some minor comments, looks good. It will be nice to update any READMEs on local setup to copy using env_template instead of example localsettings.

connectid/settings.py
hemant10yadav: I think there isn't a significant technical advantage either way - both approaches accomplish the same thing. env.db() just provides a convenience method to parse a `DATABASE_URL` string into the dictionary format Django expects

.github/workflows/ci.yaml
hemant10yadav: If we don't specify default user of `postgres` will be used.

hemant10yadav:
> Apart from some minor comments, looks good. It will be nice to update any READMEs on local setup to copy using env_template instead of example localsettings.

I will update the README as well.

connectid/settings.py
hemant10yadav: Yes it is working.

calellowitz:
Looks like tests are still failing, but otherwise looks good.

hemant10yadav:
> Looks like tests are still failing, but otherwise looks good.

I have fixed the test in [#69](https://github.com/dimagi/connect-id/pull/69) and [#70](https://github.com/dimagi/connect-id/pull/70)

sravfeyn (approved):
Great work! G2G once the test failure is addressed

calellowitz (approved):

PR #69, opened 2025-01-08
Formatting-linting

hemant10yadav:
This branch addresses formatting and linting issues.

If you haven't checked the related configuration PR yet, you can review it [here](https://github.com/dimagi/connect-id/pull/68).

I have used cmd `pre-commit run -a` to fix the issues.


messaging/migrations/0001_initial.py
pxwxnvermx, line 1: These migration files should not be affected as these are ignored from the lint and format config.

hemant10yadav: Looks like I missed adding it on `.pre-commit-config.yaml`. [da3efc](https://github.com/dimagi/connect-id/pull/68/commits/b216110f4c981b09958621655adb4e1199da3efc)

calellowitz:
Can you share how this was created? I am sure it was all automated, but just adding the command somewhere so we can see what we are reviewing.

messaging/views.py
```diff
@@ -12,24 +12,24 @@
...
 from users.models import ConnectUser
 from utils.rest_framework import ClientProtectedResourceAuth, MessagingServerAuth
 
 
 def get_current_message_server(request):
-    auth_header = request.META.get('HTTP_AUTHORIZATION')
-    encoded_credentials = auth_header.split(' ')[1]
-    decoded_credentials = base64.b64decode(encoded_credentials).decode('utf-8')
-    client_id, client_secret = decoded_credentials.split(':')
+    auth_header = request.headers.get("authorization")
```
calellowitz, line 29: do we know this is safe?

utils/__init__.py
```diff
@@ -1,42 +1,28 @@
...
-        from_=sender,
-        messaging_service_sid=settings.TWILIO_MESSAGING_SERVICE
-    )
+    client.messages.create(body=body, to=to, from_=sender, messaging_service_sid=settings.TWILIO_MESSAGING_SERVICE)
 
 
 def get_ip(request):
-    ip_address = request.META.get('HTTP_X_FORWARDED_FOR', request.META.get('REMOTE_ADDR', '127.0.0.1'))
-    return ip_address.split(',')[0]
+    ip_address = request.headers.get("x-forwarded-for", request.META.get("REMOTE_ADDR", "127.0.0.1"))
```
calellowitz, line 12: similar safety question about this, especially since the second header request is unchanged

calellowitz (commented):
Left a few questions about changes I wasn't sure of, but generally looks good.

messaging/tests.py
```diff
@@ -8,67 +10,83 @@
...
-
+from messaging.models import Channel, Message, MessageDirection, MessageStatus
+from messaging.serializers import MessageData, MessageSerializer
+from users.factories import FCMDeviceFactory
 
 APPLICATION_JSON = "application/json"
 
 
 @pytest.fixture
-def server(oauth_app):
```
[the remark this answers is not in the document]
hemant10yadav: These are test changes that I made in #70 
This change was made because earlier we had oauth app in server model but now we do not have that.

messaging/tests.py
```diff
@@ -305,37 +303,39 @@ def test_retrieve_messages_no_data(self, auth_device):
...
-        assert all(key in response_data for key in ['channels', 'messages'])
-        assert all(not response_data[key] for key in ['channels', 'messages'])
+        assert all(key in response_data for key in ["channels", "messages"])
+        assert all(not response_data[key] for key in ["channels", "messages"])
 
     def test_retrieve_messages_multiple_channels(self, auth_device, fcm_device):
         channels = ChannelFactory.create_batch(5, connect_user=fcm_device.user, server=ServerFactory.create())
         for channel in channels:
-            MessageFactory.create_batch(5, channel=channel)
+            MessageFactory.create_batch(5, channel=channel, direction=MessageDirection.MOBILE)
```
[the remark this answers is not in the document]
hemant10yadav: same updated the test in #70 acc. to the new logic.

messaging/tests.py
```diff
@@ -346,9 +346,9 @@ def test_consent(self, auth_device, channel, server, consent=False, ):
                 url=server.consent_url,
                 json_data={
                     "channel_id": str(channel.channel_id),
-                    "consent": str(consent),
+                    "consent": consent,
                 },
-                secret=server.oauth_application.client_secret
+                secret=server.secret_key,
```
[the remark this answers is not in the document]
hemant10yadav: same #70

messaging/views.py
hemant10yadav: This was automated and it was done by django-upgrade.

I checked this and from [doc](https://docs.djangoproject.com/en/4.1/ref/request-response/#django.http.HttpRequest.headers) it clearly looks safer than ever.

utils/__init__.py
hemant10yadav: This was also automated. As for your question, I changed it to how it was in older commits locally and ran pre-commit again. It changed back to the same as above, it didn't change request.META.get("REMOTE_ADDR"), only the first one was changed.

messaging/tests.py
```diff
@@ -431,5 +431,4 @@ def test_grouped_channel_messages(self, mock_send_messages, auth_device):
         data, msg_status = args
         assert isinstance(data, defaultdict) and len(data) == 2
         assert all(str(ch.channel_id) in data for ch in [channel1, channel2])
-        assert all(all(msg["received"] for msg in data[str(ch.channel_id)]["messages"]) for ch in [channel1, channel2])
```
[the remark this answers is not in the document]
hemant10yadav: I think I accidentally removed this. The only change here should be from `received` to `received_on`. I have added that [here](https://github.com/dimagi/connect-id/pull/69/commits/eeda8795ce97e6f3e220aefcb214a8ca5d78004b).

utils/__init__.py
calellowitz: is there a reason not to change the second if that is better?

PR #76, opened 2025-02-06
make sure all fields are strings

calellowitz:
This ensures all data in the FCM message is string data, which is required by firebase. Introduced in https://github.com/dimagi/connect-id/pull/75

sentry error: https://dimagi.sentry.io/issues/6193969412/?project=4508576093044736&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=14d&stream_index=2


SmittieC (approved):
Thanks!

PR #77, opened 2025-02-11
Added sentry log for 401

hemant10yadav:
Added Log401ErrorsMiddleware to capture and log 401 Unauthorised responses in Sentry. [CCCT-667](https://dimagi.atlassian.net/browse/CCCT-667)

[CCCT-667]: https://dimagi.atlassian.net/browse/CCCT-667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


utils/middleware.py
```diff
@@ -16,3 +17,22 @@ def __call__(self, request):
...
+class Log401ErrorsMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        # Check if the response status code is 401
+        if response.status_code == 401:
+            sentry_sdk.capture_message(
```
calellowitz, line 31: do you have a sense what kind of info gets included here? What from the request will we have to investigate

hemant10yadav: I tried this locally by creating a Sentry account to check what information it shows, and I saw that it normally sends things we usually see on Sentry, like the `path`, `status_code`, and `headers`. But here, it also sends the body content that comes with request. It by default filters sensitive information like passwords and tokens.

sravfeyn: If the goal of this is to debug why we are getting a 401, we should make sure that the sentry would have all that info, otherwise we should add that extra context.

sravfeyn (approved):

utils/middleware.py
```diff
@@ -16,3 +19,33 @@ def __call__(self, request):
...
+        if response.status_code == 401:
+            auth_header = request.headers.get('AUTHORIZATION', '')
+            username = None
+
+            if auth_header.startswith('Basic '):
+                try:
+                    encoded_credentials = auth_header.split(' ')[1]
+                    decoded_credentials = base64.b64decode(encoded_credentials).decode('utf-8')
+                    username, _ = decoded_credentials.split(':')
+                except Exception:
```
calellowitz, line 40: nit: this should catch a more specific exception if possible

hemant10yadav: Changed - [d33022a](https://github.com/dimagi/connect-id/pull/77/commits/d33022a755bd37edd7e00694ecc7f90f749ca71c)

calellowitz (approved):

PR #78, opened 2025-02-18
Add user counts api

pxwxnvermx:
Added the user counts API for getting the total active connect id account counts by month and year.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -1,5 +1,5 @@
 import requests
-from datetime import timedelta
+from datetime import datetime, timedelta
 from secrets import token_hex
 from urllib.parse import urlparse, urlencode
 
@@ -10,6 +10,8 @@ from django.core.exceptions import ValidationError, ObjectDoesNotExist
 from django.http import HttpResponse, JsonResponse
 from django.utils.timezone import now
 from django.views import View
+from django.db.models import Count
+from django.db.models.functions import TruncMonth
 from oauth2_provider.models import AccessToken, RefreshToken
 from oauth2_provider.views.mixins import ClientProtectedResourceMixin
 from rest_framework.decorators import api_view, permission_classes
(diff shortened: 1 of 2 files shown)


sravfeyn (approved):
Looks good, but I added some comments on CCC PR.

users/views.py
```diff
@@ -578,3 +580,19 @@ def confirm_deactivation(request):
...
         return JsonResponse({"success": True})
     return JsonResponse({"success": False})
+
+
+class FetchUserCounts(ClientProtectedResourceMixin, View):
+    required_scopes = ["user_fetch"]
+
+    def get(self, request, *args, **kwargs):
+        counts = (
+            ConnectUser.objects.filter(is_active=True)
```
calellowitz, line 590: I don't think we want to filter on active, since deactivated accounts are still accounts that have been created. Especially if we are looking at create date not current users.

users/views.py
```diff
@@ -578,3 +580,19 @@ def confirm_deactivation(request):
...
+            .annotate(date_joined_month=TruncMonth("date_joined"))
+            .values("date_joined_month")
+            .annotate(count=Count("*"))
+        )
+        count_by_year_month = {
+            str(item["date_joined_month"].date()): item["count"] for item in counts
```
calellowitz, line 596: It might be clearer to use something like `strftime` so we can be explicit about how the date is represented, otherwise it is not obvious what format it is in.

calellowitz (changes requested):
One optional, one real piece of feedback

users/views.py
pxwxnvermx: Added. https://github.com/dimagi/connect-id/pull/78/commits/daab68599f0bec5742e2188f0047f826b2b7ecd2

users/views.py
pxwxnvermx: Added. https://github.com/dimagi/connect-id/pull/78/commits/daab68599f0bec5742e2188f0047f826b2b7ecd2

calellowitz (approved):
I assume you tested this expression, if so, then good to go

PR #79, opened 2025-02-19
Delete old messages 

hemant10yadav:
This PR adds Celery periodic tasks: [CCCT-773](https://dimagi.atlassian.net/browse/CCCT-773)

- A task to delete old messages received more than 7 days ago.
- A task to resend notifications for undelivered messages.

Although we already have Celery configurations in place, I don't think we have started using it yet. I have tested it locally using Redis.

To start Celery locally, I ran: `celery -A connectid worker --loglevel=info`
I verified that Celery Beat is executing the tasks periodically using: `celery -A connectid beat --loglevel=info`

[CCCT-773]: https://dimagi.atlassian.net/browse/CCCT-773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


messaging/views.py
```diff
@@ -103,69 +102,14 @@ def post(self, request, *args, **kwargs):
...
-            message_result = send_bulk_message(message)
+            message_result = send_bulk_notification(message)
             results.append(message_result)
             if not message_result["all_success"]:
                 global_all_success = False
 
         return JsonResponse({"messages": results, "all_success": global_all_success}, status=200)
 
 
-def send_bulk_message(message):
```
hemant10yadav, line 114: Moved this function in `utils/notification.py` due to circular dependency issue.

connectid/celery_app.py
```diff
@@ -15,3 +16,14 @@
 
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
+
+app.conf.beat_schedule = {
```
pxwxnvermx, line 20: Can we handle these periodic tasks similarly to Connect? We are using `django-celery-beat` on connect for celery periodic tasks and the Tasks are saved to the database that django uses. The periodic tasks are created using django migrations.

connectid/celery_app.py
```diff
@@ -15,3 +16,14 @@
...
+    "delete_old_messages": {
```
pxwxnvermx, line 21: Do you know at what time these tasks would be run?

messaging/task.py
```diff
@@ -58,3 +63,27 @@ def send_messages_to_service_and_mark_status(channel_messages,
...
+                                               direction=MessageDirection.MOBILE).select_related('channel',
+                                                                                                 'channel__connect_user'))
+    for msg in undelivered_msgs:
+        channel = msg.channel
+        serialized_msg = MessageSerializer(msg)
+        message_to_send = NotificationData(
+            usernames=[channel.connect_user.username],
+            data=serialized_msg.data
+        )
+        send_bulk_notification(message_to_send)
```
sravfeyn, line 89: Should there be wrapped in try-except block, otherwise the rest of undelivered_msgs won't get sent too

messaging/task.py
```diff
@@ -58,3 +63,27 @@ def send_messages_to_service_and_mark_status(channel_messages,
...
     if sent_message_ids:
         Message.objects.filter(message_id__in=sent_message_ids).update(status=status_to_be_updated)
+
+
+@shared_task(name="delete_old_messages")
+def delete_old_messages():
+    """
+      Deletes messages that are older than 7 days.
+    """
+    cutoff_date = now() - timedelta(days=7)
```
sravfeyn, line 73: 7 could be declared as some sort of a constant. `MESSAGE_RETENTION_DAYS = 7`

messaging/task.py
hemant10yadav: Thanks added- [e46d420](https://github.com/dimagi/connect-id/pull/79/commits/e46d4204ed4249c1386ba673621414e973ae9357)

messaging/task.py
hemant10yadav: Thanks added- [e46d420](https://github.com/dimagi/connect-id/pull/79/commits/e46d4204ed4249c1386ba673621414e973ae9357)

connectid/celery_app.py
hemant10yadav: Celery Beat will schedule them based on when the scheduler process starts.
If you restart Celery Beat, the timers reset, meaning the tasks will be scheduled relative to the restart time.

connectid/celery_app.py
hemant10yadav: I didn't do that because the notification runs every hour, so even if Celery Beat restarts, it will still run in the next hour. The other one is just the delete message task, so we don’t care much if it gets delayed.

I thought there was no strong reason to save this in the database and make an extra db call, so that was the reason behind my choice.

messaging/task.py
```diff
@@ -58,3 +67,33 @@ def send_messages_to_service_and_mark_status(channel_messages,
...
+        channel = msg.channel
+        serialized_msg = MessageSerializer(msg)
+        username = channel.connect_user.username
+        message_to_send = NotificationData(
+            usernames=[username],
+            data=serialized_msg.data
+        )
+        try:
+            send_bulk_notification(message_to_send)
+        except Exception as e:
```
calellowitz, line 96: is this looking for a specific exception? If so it would be good to catch that, if not, it would be great if this logging went to sentry so we could access it, and the context, more easily

calellowitz (approved):
not blocking but left one suggestion

messaging/task.py
hemant10yadav: Added a log for sentry [5ef916a](https://github.com/dimagi/connect-id/pull/79/commits/5ef916a4ac42b136aadad9c62aa08ca4ae46f4ba)

calellowitz (approved):

calellowitz (approved):

PR #83, opened 2025-04-03
Send Error Response if No Recovery Phone

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-902).

This PR adds extra validation for API endpoints that use the `user.recovery_phone` field. Specifically, a new check has been added to ensure that the user has a valid recovery phone number. If it does not exist then a JSON response with an appropriate error message is returned.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -20,7 +20,7 @@ from rest_framework.views import APIView
 from utils import get_ip, get_sms_sender, send_sms
 from utils.rest_framework import ClientProtectedResourceAuth
 
-from .const import TEST_NUMBER_PREFIX
+from .const import NO_RECOVERY_PHONE_ERROR, TEST_NUMBER_PREFIX
 from .fcm_utils import create_update_device
 from .models import ConnectUser, Credential, PhoneDevice, RecoveryStatus, UserCredential, UserKey
 
@@ -85,6 +85,8 @@ def confirm_otp(request):
 @api_view(["POST"])
 def validate_secondary_phone(request):
     user = request.user
+    if not user.recovery_phone:
+        return JsonResponse({"error": NO_RECOVERY_PHONE_ERROR}, status=400)
     return PhoneDevice.send_otp_httpresponse(phone_number=user.recovery_phone, user=user)
 
 
@@ -93,6 +95,8 @@ def confirm_secondary_otp(request):
     # check otp code for user
     # mark phone as confirmed on user model
     user = request.user
+    if not user.recovery_phone:
+        return JsonResponse({"error": NO_RECOVERY_PHONE_ERROR}, status=400)
     device, _ = PhoneDevice.objects.get_or_create(phone_number=user.recovery_phone, user=user)
     data = request.data
     verified = device.verify_token(data.get("token"))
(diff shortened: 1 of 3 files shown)


pxwxnvermx (approved):

users/views.py
```diff
@@ -182,6 +188,8 @@ def confirm_secondary_recovery_otp(request):
         return HttpResponse(status=401)
     if status.step != RecoveryStatus.RecoverySteps.CONFIRM_SECONDARY:
         return HttpResponse(status=401)
+    if not user.recovery_phone:
```
calellowitz, line 191: nit: I don't think this is necessary. This code all assumes the device already exists (you can see the get call below), and only gets called when we have already sent a code to the secondary phone.

users/views.py
```diff
@@ -93,6 +95,8 @@ def confirm_secondary_otp(request):
     # check otp code for user
     # mark phone as confirmed on user model
     user = request.user
+    if not user.recovery_phone:
```
calellowitz, line 98: nit: Since this happens after the code is already sent to the phone, this is probably unnecessary, although since the next line is  `get_or_create` fine to keep.

calellowitz (approved):
This looks good. Tests would be great, but not blocking

users/views.py
zandre-eng: Removed in [4b2e845](https://github.com/dimagi/connect-id/pull/83/commits/4b2e845890501968dcf78b9e418852d5f1e3fe36).

zandre-eng:
Unit tests have been added in [46cd08d](https://github.com/dimagi/connect-id/pull/83/commits/46cd08d70c2902092c4a6ad9d2bc760fdf333b7d).

@calellowitz I struggled with getting the authentication part working for the unit tests, so please do let me know if there is a more elegant way to achieve what I've done.

calellowitz:
> Unit tests have been added in [46cd08d](https://github.com/dimagi/connect-id/pull/83/commits/46cd08d70c2902092c4a6ad9d2bc760fdf333b7d).
> 
> @calellowitz I struggled with getting the authentication part working for the unit tests, so please do let me know if there is a more elegant way to achieve what I've done.

Great question. You should be able to use the `auth_device` [fixture](https://github.com/dimagi/connect-id/blob/main/conftest.py#L26) (you can just pass it in to the test function as in [this example](https://github.com/dimagi/connect-id/blob/main/messaging/tests.py#L227) and pytest will handle creating it). That returns a client that can make authenticated requests as a user. That user doesn't have a secondary phone number so these tests should just work out of the box with that client, but if you do need to modify the user, you can pass that fixture in as well. Pytest fixtures make this stuff pretty easy, but take a bit to get used to.

zandre-eng:
> You should be able to use the auth_device [fixture](https://github.com/dimagi/connect-id/blob/main/conftest.py#L26)...

@calellowitz Ah, brilliant! I've updated the unit tests to use this fixture [546dae6](https://github.com/dimagi/connect-id/pull/83/commits/546dae6dd9ee50426cdacf8e827de7e3af160289).

calellowitz (approved):

calellowitz:
Looks great! thanks!

PR #84, opened 2025-04-03
Add Recovery Phone Field to Recovery Payload

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-913).

When a user successfully recovers their account a "recovery package" gets sent to the user's device. This PR extends the data that is included in the recovery package by adding the `recovery_phone` user field as well.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -189,7 +189,12 @@ def confirm_secondary_recovery_otp(request):
     status.step = RecoveryStatus.RecoverySteps.RESET_PASSWORD
     status.save()
     db_key = UserKey.get_or_create_key_for_user(user)
-    user_data = {"name": user.name, "username": user.username, "db_key": db_key.key}
+    user_data = {
+        "name": user.name,
+        "username": user.username,
+        "recovery_phone": user.recovery_phone,
+        "db_key": db_key.key,
+    }
     user_data.update(user_payment_profile(user))
     return JsonResponse(user_data)
 
--- a/users/views.py
+++ b/users/views.py
@@ -192,7 +192,7 @@ def confirm_secondary_recovery_otp(request):
     user_data = {
         "name": user.name,
         "username": user.username,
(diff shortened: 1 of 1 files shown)


pxwxnvermx (approved):

calellowitz (approved):
This looks great. It might be good to chat with mobile about the name of the field (it is probably fine, but they should at least know what we are calling it, and i think they generally use "secondary_phone").

calellowitz (changes requested):
I see we call it `secondary_phone` elsewhere in these views and generally send phone numbers down use `.as_e164` such as https://github.com/dimagi/connect-id/blob/main/users/views.py#L170

Given those things, I think it would be preferable to keep that consistency here

zandre-eng:
> I see we call it secondary_phone elsewhere in these views and generally send phone numbers down use .as_e164 such as
...

@calellowitz Param has been updated as suggested in [72cf491](https://github.com/dimagi/connect-id/pull/84/commits/72cf491975886aec87eef97ed5c2b2f6b55bfac3).

Let me know if you are happy with the changes.

users/views.py
```diff
@@ -320,6 +325,7 @@ def user_data(user):
     user_data = {
         "name": user.name,
         "username": user.username,
+        "secondary_phone": user.recovery_phone.as_e164 if user.recovery_phone else None,
```
calellowitz, line 328: Can you confirm with mobile (Dave) whether they need this to be an empty string or null is fine if it doesn't exist. I am fairly certain we shouldn't get into this case, but would be good to make sure mobile can handle it if we do.

calellowitz (approved):
Pending confirmation from mobile, this looks good

users/views.py
zandre-eng: I have confirmed with both Dave and Shubham from mobile regarding the above and they both said it's fine to have this be `null` if it doesn't exist.

PR #86, opened 2025-04-10
Add pytest-cov

Charl1996:
This PR adds the [pytest-cov](https://pytest-cov.readthedocs.io/en/latest/readme.html) package to check test-coverage. We can decide to keep it or not. I just thought it would be good to see how much test coverage this repo has.

See example report below run as follows: `pytest --cov`

image
 
#### Note the very last test, `utils/twilio.py`, having 42% coverage despite there being no test for it.
This is because pytest, being based on [coverage](https://coverage.readthedocs.io/en/latest/howitworks.html), runs the tests to see which lines of code has run and which not.

After adding a happy-case test, the coverage looks like this:

image

Now after adding the sad-case test the coverage has gone up to 100%, meaning all the lines in the file have been executed during testing:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 /.envrc
 *.env
 !*.env_example
+.coverage
--- a/requirements/requirements-dev.in
+++ b/requirements/requirements-dev.in
@@ -8,3 +8,4 @@ flake8
 flake8-isort
 black
 pre-commit
+pytest-cov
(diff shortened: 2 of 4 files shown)


zandre-eng (approved):
This is great! Definitely a few areas where coverage could eventually be improved (looking at you `users/views.py`).

I think it would be nice to keep this package so that we can continually assess how the coverage is looking. This package only sits in the dev requirements, so it feels quite safe to keep.

calellowitz (approved):
Looks great! thanks

PR #87, opened 2025-04-14
Add tests for confirm_recovery_pin view and handle no recovery pin set gracefully

Charl1996:
A bug was logged to sentry for the following line of code:
```
user.check_recovery_pin(recovery_pin)
```

The error:
```
object of type 'NoneType' has no len()
```

Following the code the `django.contrib.auth.hashers.check_password` package tries to identify the encoded hasher and checks the length as part of that process. If the "encoded" password is `None` (i.e. no recovery password set) this is the expected error.

The "no recovery pin" scenario is now handled more gracefully when invoking `check_recovery_pin` by raising a custom `RecoveryPinNotSetError` error and returning an appropriate message. 

### Other updates
Tests were added to test the endpoint under different conditions.


users/tests.py
```diff
@@ -194,3 +194,71 @@ def test_fetch_credential_with_org_slug(self, authed_client):
     def test_fetch_credential_without_org_slug(self, authed_client):
         response = authed_client.get(self.url)
         self.assert_statements(response, expected_count=13)
+
+
+@pytest.fixture
+def recovering_user(user):
```
zandre-eng, line 200: I think moving this to `conftest.py` might be a better fit, as that is where the other fixtures are located.

users/tests.py
```diff
@@ -194,3 +194,71 @@ def test_fetch_credential_with_org_slug(self, authed_client):
...
+
+@pytest.mark.django_db
+class TestRecoveryPinConfirmationApi:
+    url = reverse("confirm_recovery_pin")
+
+    def _get_recovery_status(self, user):
+        return RecoveryStatus.objects.get(user=user)
+
+    def test_recovery_status_secret_mismatch(self, recovering_user, client):
+        self._get_recovery_status(recovering_user)
```
zandre-eng, line 215: Instead of passing in `recovering_user` and then still needing to fetch the recovery status, how about instead creating a fixture that returns the recovery status? Doing it this way you'll have access to the status and can then also access the user via `recovery_status.user`.

users/tests.py
```diff
@@ -194,3 +194,71 @@ def test_fetch_credential_with_org_slug(self, authed_client):
     def test_fetch_credential_without_org_slug(self, authed_client):
         response = authed_client.get(self.url)
         self.assert_statements(response, expected_count=13)
+
+
+@pytest.fixture
+def recovering_user(user):
+    RecoveryStatusFactory(user=user, step=RecoveryStatus.RecoverySteps.CONFIRM_SECONDARY)
```
zandre-eng, line 201: I'm not sure you need to pass in the `user` here. Looking at `RecoveryStatusFactory` it is already creating a new user with `user = factory.SubFactory(UserFactory)`

users/tests.py
```diff
@@ -194,3 +194,71 @@ def test_fetch_credential_with_org_slug(self, authed_client):
...
+    user.set_recovery_pin("1234")
```
zandre-eng, line 202: Instead of setting this here, do you think it would be possible to add this to the `UserFactory` class directly instead? That way you don't have to do a separate DB call to set this property after instantiating the user.

users/views.py
```diff
@@ -368,8 +369,13 @@ def confirm_recovery_pin(request):
     if status.step != RecoveryStatus.RecoverySteps.CONFIRM_SECONDARY:
         return HttpResponse(status=401)
     recovery_pin = data["recovery_pin"]
-    if not user.check_recovery_pin(recovery_pin):
-        return JsonResponse({"error": "Recovery PIN is incorrect"}, status=401)
+
+    try:
+        if not user.check_recovery_pin(recovery_pin):
+            return JsonResponse({"error": "Recovery PIN is incorrect"}, status=401)
```
zandre-eng, line 375: Just noting that in [this](https://github.com/dimagi/connect-id/pull/85/commits/cf1a116b8e766666cce8589a09e9df3f688ade35) PR I've added a class so that we can start adding error status codes to the JSON response instead of a string message. This should allow more flexibility for mobile in being able to provide a localised message using the status.

It might be worth using a status code here as well for consistency's sake.

zandre-eng (commented):
Looks good, just a few small comments from my side.

users/tests.py
Charl1996: Created a conftest.py file for users and moved it there. See https://github.com/dimagi/connect-id/pull/87/commits/5e0e5a44a4d6a024bb0f2b90f9542fe722aa4476.

users/tests.py
zandre-eng: That's a fair point. Fine to leave as is then :+1:

zandre-eng (commented):
Thanks for the changes @Charl1996, looks good! Happy to approve once the status code changes have been implemented.

users/views.py
```diff
@@ -368,8 +369,13 @@ def confirm_recovery_pin(request):
...
+    except RecoveryPinNotSetError as e:
+        return JsonResponse({"error": str(e)}, status=401)
```
calellowitz, line 377: nit: I think this should be a 400. We generally use that for these "impossible" requests. It isn't that the PIN is wrong, its that they shouldn't be asking for a PIN in the first place

calellowitz (approved):
Agree on Zandre's comment about incorporating his error code model. Otherwise one non-blocking preference, but definitely feel free to push back on that if you have a strong preference

users/views.py
Charl1996: Ah, returning a 400 makes more sense, yes.

Charl1996:
@zandre-eng Now that we're using the ErrorCodes, will we systematically update all the error responses to add error codes? If so, do we know if mobile is compatible with the new API responses?

zandre-eng:
> Now that we're using the ErrorCodes, will we systematically update all the error responses to add error codes? If so, do we know if mobile is compatible with the new API responses?

@Charl1996 I'm not sure how the rollout will be on mobile for the new error codes, just that it was recommended we start implementing them as a consistent standard across the API endpoints. To that extent, it should be compatible in the sense of adding the error code to the response won't break anything, however I think we should take this on a case-to-case basis. If we are making change to an API endpoint's response, it's a good idea to just double check with mobile and confirm whether we can safely change what is getting returned.

Charl1996:
I've updated the "incorrect code" error message format also for consistency. I've verified that only the status code is read currently on mobile, so it's a safe change.

zandre-eng (approved):
Thanks for the updated responses, looks good!

PR #88, opened 2025-04-14
Add Deactivation Tokens for Demo Users

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-940).

This PR tweaks the endpoint for fetching demo users with the following changes:
- All users with a valid deactivation token are included in the list of demo users.
- Only phone device instances with a valid OTP will now be included in the query.

Additional unit tests have been created to verify that the demo users endpoint is fetching and sorting the data correctly.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from django.contrib.auth.hashers import check_password
 from django.contrib.auth.password_validation import validate_password
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
-from django.db.models import Count
+from django.db.models import Count, F
 from django.db.models.functions import TruncMonth
 from django.http import HttpResponse, JsonResponse
 from django.utils.timezone import now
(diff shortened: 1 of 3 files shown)


pxwxnvermx (approved):
Thanks 🙌🏽

users/views.py
```diff
@@ -411,10 +411,25 @@ class GetDemoUsers(ClientProtectedResourceMixin, View):
...
+        demo_connect_users = (
+            ConnectUser.objects.filter(
+                phone_number__startswith=TEST_NUMBER_PREFIX,
+                deactivation_token__isnull=False,
+            )
+            .annotate(token=F("deactivation_token"))
+            .values("phone_number", "token")
+        )
+
+        demo_users = list(demo_phone_devices) + list(demo_connect_users)
```
Charl1996, line 427: @zandre-eng Do we care about duplicates in the `demo_users` resultant list?

Charl1996: > Do you mean duplicates in the sense of there might be an entry in both demo_phone_devices and demo_connect_users 

Yes.

Ok, cool. Seems like duplicates are fine then.

Charl1996 (approved):

PR #89, opened 2025-04-16
Resolve Exception Where Message instance Already Exists

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-983).
Link to Sentry error [here](https://dimagi.sentry.io/issues/6194505694/?project=4508576093044736&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=3).

This PR resolves an unhandled exception that has been reported on Sentry. Specifically, making a POST request to the `SendMobileConnectMessage` endpoint with message IDs that already exist causes an unhandled `IntegrityError` exception to get raised. This is because we attempt to bulk create new `Message` instances with the given message IDs, regardless of whether the messages already exist or not.

Now, if there are existing messages for any of the given message IDs then these will simply be skipped in the creation process. The existing messages will however, still have their status checked to ensure that they have already been sent. If not, they will be sent along with any new messages.

### DIFF
--- a/messaging/serializers.py
+++ b/messaging/serializers.py
@@ -38,6 +38,7 @@ class BulkMessageSerializer(serializers.Serializer):
 
 class MessageSerializer(serializers.ModelSerializer):
     ciphertext = serializers.SerializerMethodField()
+    channel = serializers.SerializerMethodField()
     tag = serializers.SerializerMethodField()
     nonce = serializers.SerializerMethodField()
     message_id = serializers.SerializerMethodField()
@@ -61,3 +62,6 @@ class MessageSerializer(serializers.ModelSerializer):
 
     def get_action(self, obj):
         return CCC_MESSAGE_ACTION
+
+    def get_channel(self, obj):
+        return str(obj.channel_id)
--- a/messaging/tests.py
+++ b/messaging/tests.py
@@ -168,8 +168,7 @@ class TestCreateChannelView:
 
         return response
 
-    def test_create_channel_success(self, client, fcm_device, oauth_app):
-        server = ServerFactory.create()
+    def test_create_channel_success(self, client, fcm_device, oauth_app, server):
         data = rest_channel_data(fcm_device.user)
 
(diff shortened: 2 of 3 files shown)


messaging/serializers.py
```diff
@@ -61,3 +62,6 @@ def get_message_id(self, obj):
 
     def get_action(self, obj):
         return CCC_MESSAGE_ACTION
+
+    def get_channel(self, obj):
+        return str(obj.channel_id)
```
zandre-eng, line 67: Trying to serialize existing `Message` instances in the `SendMobileConnectMessage` class results in the `channel` field being a `UUID` object instead of a `str`, which is what we get from `Message` instances returned with `bulk_create()`.

Adding this to the serializer is simply to enforce consistency and ensure we always have the channel ID in `str` format.

messaging/tests.py
```diff
@@ -272,8 +272,30 @@ def test_multiple_messages(self, auth_device, channel, server):
...
+                data=json.dumps(data),
+                content_type=APPLICATION_JSON,
+            )
+            json_data = response.json()
+            expected_message_data = self._get_expected_message_data([pending_msg], channel, server)
+            mock_send_bulk_message.assert_called_once()
             mock_send_bulk_message.assert_called_once_with(expected_message_data, MessageStatus.SENT_TO_SERVICE)
 
+        assert response.status_code == status.HTTP_201_CREATED
+        assert json_data["message_id"] == [pending_msg.message_id]
```
Charl1996, line 297: So when a duplicate message has a status of `PENDING`, we also expect only one to be created but for both to be sent, right?

zandre-eng: If we have an already existing message we'll skip trying to re-create it, however we will still attempt to send it again if it has a `PENDING` status.

So in the case of this test we have two messages that already exist, but one has a status of `PENDING` and the other `SENT`. We won't attempt to recreate either, and only the message with the `PENDING` status will get sent again since it is still waiting to be sent.

Charl1996 (approved):

PR #91, opened 2025-04-24
Service for checking app integrity with google

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-1073)

This PR adds a class which validates a given google play integrity token. If the token verdict (from google) does not meet certain criteria the class will raise an appropriate error. It's necessary to know which error was raised in order to be able to return a unique response to the mobile client, since there's certain remedies that the mobile user can perform based on the error.

The class will be used in an upcoming PR as part of a decorator which will be applied to views where the token verification will be needed before the request can be completed.

**Note**
The google integration needs to be tested still once we have the appropriate google credentials json file. Google's documentation is rather scattered on exactly how to interact with the play integrity api. Between [some documentation](https://github.com/googleapis/google-api-python-client/blob/main/docs/start.md#building-and-calling-a-service) and an [example repo](https://github.com/yvestumushimire/play-integrity-python/blob/241afcdc9951fe418bd00fa849876fe4f8953243/play_integrity/integrity.py#L30) I've managed to put together the code in `_obtain_verdict`.


zandre-eng (approved):
Looks good! Just a few minor comments.

utils/app_integrity/google_play_integrity.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class AppIntegrityService:
+    """
+    Verifies the application integrity of the app using Google Play Integrity API.
+    """
+
+    def __init__(self, token: str, request_hash: str = None):
+        self.token = token
+        self.request_hash = request_hash
```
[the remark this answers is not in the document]
Charl1996: Ah, that's legacy code. Will update. Not sure actually why I initially did that.

utils/app_integrity/google_play_integrity.py
```diff
@@ -0,0 +1,85 @@
...
+        with build(**service_spec) as service:
+            body = {"integrityToken": self.token}
+            response = service.v1().decodeIntegrityToken(packageName=APP_PACKAGE_NAME, body=body).execute()
+
+        return VerdictResponse.from_dict(**response)
+
+    @property
+    def _google_service_account_credentials(self):
+        credentials, _ = google.auth.load_credentials_from_file(
+            settings.GOOGLE_APPLICATION_CREDENTIALS, scopes=[AUTH_SCOPE]
```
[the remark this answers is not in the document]
Charl1996: Not a bad idea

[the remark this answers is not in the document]
Charl1996: https://github.com/dimagi/connect-id/pull/91/commits/08f7f77eb0c68a99f7cff89a25d9027219fd894c

zandre-eng (approved):

calellowitz (approved):

utils/app_integrity/google_play_integrity.py
```diff
@@ -45,18 +46,19 @@ def _obtain_verdict(self) -> VerdictResponse:
...
     @property
-    def _google_service_account_credentials(self):
+    def _google_service_account_credentials(self) -> Credentials:
         if not settings.GOOGLE_APPLICATION_CREDENTIALS:
             raise Exception("GOOGLE_APPLICATION_CREDENTIALS must be set")
 
-        credentials, _ = google.auth.load_credentials_from_file(
-            settings.GOOGLE_APPLICATION_CREDENTIALS, scopes=[AUTH_SCOPE]
-        )
+        credentials, _ = google.auth.load_credentials_from_file(settings.GOOGLE_APPLICATION_CREDENTIALS, scopes=[])
```
Charl1996, line 61: I removed the scope because there's some permission issue otherwise where the credentials provided is not authorized for that scope. Truth be told I'm not 100% sure that the provided scope is the right one for play integrity (docs are not necessarily very helpful), so I'm leaving it for now.

Charl1996:
I've tested the request to google with real credentials but a fake token and got a response where google basically says "this token is not valid". I have confidence that we can at least talk to google at this point and that the responses can be handled appropriately (the `.json` responses are from the google play console).

utils/app_integrity/google_play_integrity.py
```diff
@@ -80,7 +80,7 @@ def _check_app_integrity(self, app_integrity: AppIntegrity):
             raise AppIntegrityError("App not recognized")
 
     def _check_device_integrity(self, device_integrity: DeviceIntegrity):
-        if device_integrity.deviceRecognitionVerdict != "MEETS_DEVICE_INTEGRITY":
+        if device_integrity.deviceRecognitionVerdict[0] != "MEETS_DEVICE_INTEGRITY":
```
zandre-eng, line 85: `deviceRecognitionVerdict` is defined as a `str` in the `DeviceIntegrity` class, so confirming whether that will need to change to match using it as a list here?

Also, do we know if this index will always exist? Can there be a scenario where it's simply an empty list?

Charl1996: Oh, good catch! Let me update.

> do we know if this index will always exist?

I'm fairly confident that the answer is yes.

Charl1996: https://github.com/dimagi/connect-id/pull/91/commits/1c9db5223d7f96357a1dbd87cd04bb0a2079f740

zandre-eng (approved):

PR #92, opened 2025-04-24
Integrity check required decorator

Charl1996:
Builds on [this PR](https://github.com/dimagi/connect-id/pull/91).
[Ticket](https://dimagi.atlassian.net/browse/CCCT-1074)

This PR adds a decorator which makes use of the `AppIntegrityService` to verify the request comes from a legit mobile device and legit app by pulling the required information **from the request header**. The decorator returns an appropriate response to the mobile app when the verification fails. 

This PR also introduces a new API version, `2.0` (which the decorator checks). The default API version is also set to `v2`.

#### Why `V2.0` and not a minor version of `V1`?
Because it's a breaking change which is not backwards compatible.

Appropriate tests have been created for the view(s) being decorated at this time. 

(Looping in @shubham1g5 just as an FYI)


shubham1g5:
> We should coordinate with the mobile team on when that can happen as they need to make sure to send the appropriate data up to the server first, lest the existing endpoints would start producing errors

Think whenever we do this, there will still be possibility of old app versions out there to not send that data. Hence we will need to take a decision here whether to error the old builds out vs implement some sort of api versioning to only error for newer api versions.

utils/app_integrity/decorators.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Checks the integrity of the app using the Google Play Integrity API.
+    """
+
+    def wrapper(request, *args, **kwargs):
+        integrity_token = request.POST.get("integrity_token")
+        request_hash = request.POST.get("request_hash")
+
+        if not (integrity_token and request_hash):
+            return JsonResponse({"error_code": "INTEGRITY_DATA_MISSING"}, status=400)
```
zandre-eng, line 22: Could we do something similar to what we did for the `users` app where we had an `ErrorCodes` class holding const strings of all the possible error codes?

Charl1996: https://github.com/dimagi/connect-id/pull/92/commits/15aad1f160ff85fedaf71d94cf253013d19ba0bd

Charl1996:
> Think whenever we do this, there will still be possibility of old app versions out there to not send that data. Hence we will need to take a decision here whether to error the old builds out vs implement some sort of api versioning to only error for newer api versions.

That's a good idea. I'll start some discussions.

zandre-eng (approved):

Charl1996:
> we are creating a security mechanism but allowing people to opt out of it (by specifying an old version)

@shubham1g5 Can you confirm whether mobile does in fact specify version `1.0` in the headers currently? If so, then I think we can make the default version `v2.0` on the backend, meaning it expects `v2.0` if nothing is specified. I just don't want to update the default if none of the requests have a version specified and we roll this out, in which case we'll break the register endpoint for all apps not having updated to the latest code (assuming there's a newer app supporting v2).

connectid/settings.py
```diff
@@ -181,8 +187,11 @@
     ],
     "DEFAULT_THROTTLE_RATES": {"anon": "100/day", "user": "1000/day"},
     "DEFAULT_VERSIONING_CLASS": "rest_framework.versioning.AcceptHeaderVersioning",
-    "DEFAULT_VERSION": "1.0",
-    "ALLOWED_VERSIONS": ["1.0"],
+    "DEFAULT_VERSION": APIVersion.V1,
```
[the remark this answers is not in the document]
Charl1996: Awaiting @shubham1g5 's reply.

users/views.py
```diff
@@ -26,9 +27,9 @@
 from .models import ConnectUser, Credential, PhoneDevice, RecoveryStatus, UserCredential, UserKey
 
 
-# Create your views here.
 @api_view(["POST"])
 @permission_classes([])
+@require_app_integrity
```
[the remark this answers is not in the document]
Charl1996: If I run the tests with `v3.0` it fails with `AssertionError: b'{"detail":"Invalid version in \\"Accept\\" header."}'` before the `require_app_integrity` decorator executes, so the the version is checked before my decorator even executes.

Checking the `api_view` decorator I can confirm that the version is indeed checked before the decorated function is executed.

shubham1g5:
> Can you confirm whether mobile does in fact specify version 1.0 in the headers currently?

I am not 100% sure, @OrangeAndGreen can you confirm ?

Charl1996:
Confirmed with @OrangeAndGreen in person that headers specify api version, hence bumping default api version to `v2`. From now on, no mercy....

Charl1996:
I've added a lot of tests, particularly pertaining to the views requiring an app check. Our test coverage sits at 89% now.

users/views.py
```diff
@@ -318,9 +330,14 @@ def update_profile(request):
 
 
 @api_view(["POST"])
+@require_app_integrity
 def set_recovery_pin(request):
     data = request.data
     user = request.user
+
+    if not data.get("recovery_pin"):
+        return JsonResponse({"error": "Missing recovery pin"}, status=400)
```
zandre-eng, line 339: Nit: Could we make these error responses consistent with our other recent changes by making it a status code (e.g. `MISSING_RECOVERY_PIN`)?

zandre-eng (approved):
Looks good. Just confirming whether the `reset_password` endpoint also need the integrity decorator added?

zandre-eng (approved):
Thanks!

Charl1996:
@zandre-eng Apparently mobile will only start with the 2 endpoints for now and will bump the API version up to `v2.0`. However, since bumping it means all requests will now sit on `v2.0`, we need to make sure the server only decorates the 2 views which mobile is prepared to send tokens for, lest the server breaks the rest of the endpoints.

I've thus removed the token check again from all but 2 endpoints.

zandre-eng (approved):

Charl1996:
Once again some important changes...

Since we decided that we're only going to do the integrity check on the "start configuration" endpoint (which does not exist yet), I've removed the integrity check from all endpoints here.

This PR is now reduced to doing two things:
1. Create the integrity check decorator
2. Add some more unit tests

zandre-eng (approved):

PR #93, opened 2025-04-24
Catch Exception Where Channel Does Not Exist

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-984).

This is a small change that simply catches the exception caused by the `SendServerConnectMessage` class when it attempts to create a `Message` instance with a channel that does not exist. 

The most probable reason as to why the channel wouldn't exist is that an admin user deleted the channel in ConnectID and a reference to the channel still exists in CommCare HQ.


messaging/views.py
```diff
@@ -210,15 +212,18 @@ def post(self, request, *args, **kwargs):
         content = data["content"]
         for field in ("nonce", "tag", "ciphertext"):
             if not content[field]:
-                return JsonResponse({"errors": "invalid message content"}, status=status.HTTP_400_BAD_REQUEST)
+                return JsonResponse({"errors": ErrorCodes.INVALID_MESSAGE_CONTENT}, status=status.HTTP_400_BAD_REQUEST)
         message_data = {
             "channel_id": data["channel"],
```
Charl1996, line 217: Just so I understand, this channel id comes from HQ?

PR #95, opened 2025-05-01
Additional Data for Push Notifications

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-990).

This PR adds the following for outbound push notifications for mobile:
- A notification title and body have been added.
- The channel name is now included in the data payload for the notification.

The goal with the above changes are to help prevent push notifications from failing to appear on the user's mobile device.

### DIFF
--- a/messaging/tests.py
+++ b/messaging/tests.py
@@ -218,7 +218,12 @@ def test_send_fcm_notification_view(client, channel):
 
         serialized_msg = MessageSerializer(db_msg).data
         serialized_msg["channel"] = str(db_msg.channel.channel_id)
-        expected = MessageData(usernames=[channel.connect_user.username], data=serialized_msg)
+        expected = MessageData(
+            usernames=[channel.connect_user.username],
+            data=serialized_msg,
+            title="New Connect Message",
+            body=f"You received a new message from {channel.visible_name}",
+        )
 
         mock_send_bulk_message.assert_called_once_with(expected)
 


messaging/views.py
```diff
@@ -236,6 +236,8 @@ def post(self, request, *args, **kwargs):
         message_to_send = MessageData(
             usernames=[channel.connect_user.username],
             data=MessageSerializer(message).data,
+            title="New Connect Message",
+            body=f"You received a new message from {channel.visible_name}",
```
zandre-eng, line 240: Curious to hear from others whether the above title and body strings are fine, or whether folks perhaps have suggestions on something better we can put in here for the notification?

Charl1996:
@zandre-eng 

> The goal with the above changes are to help prevent push notifications from failing to appear on the user's mobile device.

Why is this? Is it an android thing?

messaging/views.py
Charl1996: What would be an example of the channel's visible name?

Charl1996 (approved):
No concerns from my side

zandre-eng:
> Why is this? Is it an android thing?

@Charl1996 Yeah, based on what I understand from the related Jira tickets it sounds like if the notification doesn't contain a title and body then it won't pop up in the user's system tray if the app is not currently open.

messaging/views.py
zandre-eng: For context this currently defaults to be the domain name, however we are able to override it in the HQ project settings to be any name. I'm assuming this is the case when the domain name isn't super recognisable to mobile workers.

calellowitz (approved):
looks good to me

PR #96, opened 2025-05-08
kamal deploy

calellowitz:
This heavily borrows from https://github.com/dimagi/connect-id/pull/17 and our commcare connect setup to allow deploy of connectid via kamal. I plan to roll this out tomorrow, but wanted to open the PR for review. The two commits handle the docker setup and then the deploy configuration. All IPs referece a new elastic IP that I provisioned, but have not attached yet to an EC2 instance. It will be used for the connectid web server when we roll this out.

Settings changes were either required for the new docker setup or to bring our codebase in line with updated localsettings in production. Once this is deployed, I will make sure the deployed env file matches the one in 1password, and all future settings should be made there.


docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    environment:
+      - POSTGRES_USER=connect
+      - POSTGRES_PASSWORD=connect
+      - POSTGRES_DB=connect
+    ports:
+      - '5433:5432'
+    healthcheck:
+      test: pg_isready -d $${POSTGRES_DB} -U $${POSTGRES_USER}
+      interval: 2s
+      retries: 10
```
pxwxnvermx, line 28: We need to add redis service in the docker-compose for celery.

README.md
```diff
@@ -43,3 +41,51 @@
...
+## Production Deploy
+
+### Setup
+
+#### Kamal
+
+(requires Ruby)
+
+```bash
+gem install kamal -v '~> 1.0.0'
```
pxwxnvermx, line 54: nit: We can use the latest version v1 (v1.9.2) of Kamal here instead of v1.0.0

README.md
```diff
@@ -43,3 +41,51 @@
...
+```bash
+aws configure sso --profile commcare-connect
+aws sso login --profile commcare-connect
+```
+
+Note: If you used a different profile name you will need to set the `AWS_PROFILE` environment variable to the profile name.
+
+### Deploy
+
+To deploy run `kamal deploy` from within the `deploy` directory
```
Charl1996, line 91: @calellowitz 
Correct me if I'm wrong (I'm not an expert on these configurations), but it seems like when running the deploy command the code on my machine is deployed and not the remote `main` branch? Is the expectation that the developer deploying will update their local `main` with remote `main`?

README.md
calellowitz: Great catch

docker-compose.yml
calellowitz: Also a good catch, and will remove the web container we talked about, since I agree its not really necessary

README.md
calellowitz: That is correct, the process would be to checkout `main` locally and pull latest code from github before deploying. V2 of this work will be to get it working with github actions, so that becomes unnecessary, but I wanted to get this out ASAP since it was such a big improvement over existing workflows

Charl1996 (approved):
Looks good to me

calellowitz:
merging since the last commit is docs only and the rest is approved

PR #97, opened 2025-05-08
Upload User Photos to S3

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1152).

This PR adds a new mechanism in which user photos will be uploaded to an S3 bucket for storage. This makes use of a new required `boto3` library which is needed to make the necessary request to S3 to upload the image data.

Testing the S3 upload changes was done locally using a dockerised version of Minio which is able to mimick S3.

Additionally, the `update_profile` endpoint has been updated to allow for updating the user's photo.


users/models.py
```diff
@@ -35,6 +35,7 @@ class ConnectUser(AbstractUser):
     recovery_phone_validation_deadline = models.DateField(blank=True, null=True)
     deactivation_token = models.CharField(max_length=25, blank=True, null=True)
     deactivation_token_valid_until = models.DateTimeField(blank=True, null=True)
+    photo = models.TextField(blank=True, null=True, help_text="Base64 encoded image")
```
zandre-eng, line 38: Since base64 encoded strings can be of varying lengths (e.g. 270k characters for a 200kb image) I opted to make this a text field without a max length attribute.

calellowitz (changes requested):
The view seems fine, but I am curious about some of the choices made here for storing the image. In general storing large objects like this inline in postgres isn't great. It dramatically increases row size, which could cause performance issues both at the DB level (where there is a lot more disk i/o and memory required to read and write these rows), and at the django level where these photos will be pulled into memory when users are fetched. The user table in particular is the most used and most critical table in connectid so we should be careful in making changes that could have large impacts.

A few questions:
1) Was S3 or other dedicated object storage considered? That would eliminate the postgres disk and performance concerns, but potentially at the cost of slower fetching (especially if we needed to fetch in bulk, but that doesn't seem like a real use case).
2) Why store in base64 instead of binary? I think this would reduce the size by about 25%
3) If we do decide to store in postgres, was a dedicated table for these considered with a one to one foreign key to the user? That way queries to the user table not involving the photo would stay quite small, but it would be easy to access as necessary

Charl1996:
> we are simply saving the base64 encoded image strings and not the image files themselves

Should we be storing the image files? What's the difference?

zandre-eng:
> Should we be storing the image files? What's the difference?

@Charl1996 If we're going to go ahead and store the images on S3 then storing the image files instead of the base64 string would make more sense.

From what I understand base64 is the easiest way to send files when using JSON, so we'll be receiving the base64 encoded string of the user's photo from mobile. From here, we can simply convert it back to an image file on the server and then upload it to S3.

Charl1996:
> What's the difference?

For what it's worth, I ChatGPT'ed this question. Turns out storing base64 could actually inflate the file size by ~33% due to the fact that the encoding is being represented and stored as ASCII characters. This in turn can have an influence on performance.  

The long and short is that storing the binary file is always recommended unless there's a specific reason not to do that.

zandre-eng:
Apologies for the force push, but given the drastic change in direction I thought it best to drop the previous commits.

zandre-eng:
An open question that I have for folks reviewing this PR:

I used Minio for doing local testing of the S3 code to check that it runs as expected and uploads data correctly to a imitated S3 bucket. The automated tests I've created do not include using Minio for testing - is this something that we would like to include?

My initial hestitation for saying yes is that this means any dev running tests for this locally will need to have a dockerised version of Minio running. While doable, this does mean we'll have to maintain an additional dependency for what is currently a single unit test. This felt a bit overkill for the sake of a single test, though I am curious to hear if folks feel differently.

FYI @calellowitz @Charl1996

connectid/settings.py
```diff
@@ -291,3 +291,7 @@
 
 
 GOOGLE_APPLICATION_CREDENTIALS = env("GOOGLE_APPLICATION_CREDENTIALS", default="")
+
+
+# Bucket where user photos are stored on S3
+AWS_S3_PHOTO_BUCKET_NAME = env("AWS_S3_PHOTO_BUCKET_NAME", default="personalid-user-photos")
```
zandre-eng, line 297: Would we prefer to rather not specify a default here and have it always be explicitely given in the `.env` settings?

users/services.py
```diff
@@ -0,0 +1,20 @@
+import base64
+
+import boto3
+from django.conf import settings
+
+
+def upload_photo_to_s3(image_base64, user_id):
+    filename = f"{user_id}.jpg"
+    s3_client = boto3.client("s3")
```
zandre-eng, line 9: No credentials are being passed here as this code expects to be running inside an EC2 instance where an IAM role with S3 access has been linked to the EC2 instance.

connectid/settings.py
Charl1996: Unless the bucket name is private I don't see why we can't specify a default.

zandre-eng: Bucket names are public in the sense of they can be probed and discovered by an outside source using the correct methods. As long as the name itself doesn't hold any sensitive information (e.g. an ID) then I don't see any concern with having it be visible here. Instead, the importance is in ensuring that proper access control is setup for the bucket.

Charl1996:
@zandre-eng 

> The automated tests I've created do not include using Minio for testing - is this something that we would like to include?

I think what you've done is perfectly fine.
```
        mock_s3.put_object.assert_called_once()
        _, kwargs = mock_s3.put_object.call_args
        assert kwargs["Key"] == f"{user.id}.jpg"
``` 
Adding minio will only allow us to additionally retrieve the image to verify that it has indeed been stored, but that's then testing to make sure the library works, which I think is beyond the scope of this app and what we have to include in a test. Making sure the `put_object` method works is an expectation that rests on those maintaining the library.

Or is there a different reason you were thinking of for adding minio?

Charl1996 (approved):
Thanks for the changes. Looks good to me.

zandre-eng:
> Adding minio will only allow us to additionally retrieve the image to verify that it has indeed been stored, but that's then testing to make sure the library works, which I think is beyond the scope of this app and what we have to include in a test. Making sure the put_object method works is an expectation that rests on those maintaining the library.

@Charl1996 I agree with this, and is what I was thinking as well.

> Or is there a different reason you were thinking of for adding minio?

The only benefit I could perhaps see with having Minio for testing is that it could help ensure that we are passing the correct arguments to `put_object` such that it doesn't raise an unexpected exception, or that the format of our image data gets correctly uploaded to the bucket without any issues. In saying that, I don't believe these warrant the additional overhead and the function call doesn't have input arguments that complex that would require special testing.

users/services.py
```diff
@@ -0,0 +1,20 @@
...
+    try:
+        image_data = base64.b64decode(image_base64)
+        s3_client.put_object(
+            Bucket=settings.AWS_S3_PHOTO_BUCKET_NAME,
+            Key=filename,
+            Body=image_data,
+            ContentType="image/jpeg",
```
calellowitz, line 16: do we know its always a jpeg?

users/const.py
```diff
@@ -3,6 +3,9 @@
 NO_RECOVERY_PHONE_ERROR = "Recovery phone number does not exist for user"
 
 
+MAX_PHOTO_SIZE = 1 * 1024 * 1024  # 1 MB
```
calellowitz, line 6: Does mobile know about this limit and handle it in a useful way for the user?

users/services.py
```diff
@@ -0,0 +1,20 @@
...
+        )
+    except Exception:
```
calellowitz, line 18: This will silence any error so we won't know they are occurring. If there are certain known exceptions we want to ignore, we should catch those explicitly. Otherwise, we should propagate the errors to sentry at a minimum and are potentially OK failing hard if they occur, especially since we will return a failing status code to the mobile in that case anyway.

users/tests.py
```diff
@@ -378,3 +378,35 @@ def test_confirm_recovery_pin_success(self, recovery_status, client):
...
+@pytest.mark.django_db
+class TestUpdateProfile:
+    test_number = "+27734567657"
+
+    url = reverse("update_profile")
+
+    def test_success(self, auth_device, user):
+        data = {"name": "FooBar", "photo": "123456789"}
+        user.phone_number = self.test_number
+        user.save()
```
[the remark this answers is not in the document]
calellowitz: The easiest answer is probably to hard code the phone number in the factory. We don't need to randomize using faker so having one with a valid number set up may be fine for 90% of use cases.

users/services.py
zandre-eng: For the standard Android app camera this should always be the case. But thinking about it now, users can still use a third-party camera app however, so it might be good to cover this edge case and return a 400 back to mobile if the format is incorrect. I'll update  this.

users/const.py
zandre-eng: Yes, Shubham was the one that requested we put in a size limit. I let him know that its a 1mb limit on the Jira ticket itself.

users/services.py
zandre-eng: I'd like to keep it to a caught exception (as opposed to failing hard) so that we can still return a helpful status code to mobile, however I do agree that this should bubble up to Sentry. 

I've made some tweaks to capture this exception to Sentry ([2883be5](https://github.com/dimagi/connect-id/pull/97/commits/2883be54ce7bb41c935a9e505479729706979dce)).

users/services.py
zandre-eng: I've added an additional check to ensure that the given base64 string is a JPEG image ([02ff4cc](https://github.com/dimagi/connect-id/pull/97/commits/02ff4ccbc95c6e17dbd535d0dd7b634e6c4c9641)).

users/const.py
calellowitz: Thanks, I just want to make sure if we are erroring here, the phone knows why, and can handle it

users/services.py
calellowitz: > we can still return a helpful status code to mobile
Can you say more about this? Specifically the ways this is more helpful. I would have imagined that a generic "this failed" with a 400 error would not have more info than  a 500 (both convey that the request failed), and a 500 seemed slightly more correct, since it represents an unexpected server error rather than the phone having made a bad request. If we could give the phone useful info to resolve the situation, I agree, but since we don't actually know what went wrong, a 500 feels as right (and 5xx implying server error rather than 4xx implying client error).

This solution addresses the monitoring issue, but doesn't seem better than the simpler version, unless we wanted the rest of the code in the view to execute.

users/services.py
calellowitz: A better way to ask the original question would be whether we care. I am not sure how we are using these image formats, but I imagine we should be content with any image format. If a phone sends a png instead of a jpeg, I would think we can just save it. I was more suggesting allowing other formats (and probably having the phone specify the format in the request). However, if this is easier, or mobile can enforce it, this is probably fine.

users/services.py
zandre-eng: > Can you say more about this?

What I mean by helpful is that by returning a predictable/static error code (as opposed to leaving the exception uncaught), mobile can then use this error code to present a translated helpful message to the user based on what code was received. This feels slightly more helpful than having the uncaught exception simply respond with the server error page template. Additionally, this also feels more consistent and in line with the other changes we've been making on ConnectID to always returned a predictable error code if something went wrong.

In saying the above, that was moreso focused around catching the exception rather than leaving it uncaught. Regarding your thought to have the response be a 500 instead of a 400, I am happy to make this change and can see your point on how this could be considered a server error rather than a client error.

users/services.py
zandre-eng: >  If a phone sends a png instead of a jpeg, I would think we can just save it

I agree this makes sense if mobile can send what type the photo is in the request. Otherwise, this could become a burdensome task of trying to figure out what extension the base64 string is so that we can add the correct extension to the filename. Given that jpeg is one of the most common extensions I thought it easier to simply enforce a specific format rather than potentially leaving ourselves open to edge cases by supporting any format.

But let me check with mobile whether they'll be able to send the extension type. If they can, then we should be able to easily support saving any file type. I'll also like to check whether mobile requires/expects a certain file type to be able to show it correctly on the phone, or if any file type can be shown. If it's the former then adding an extension restriction makes more sense.

zandre-eng: I've received an update from mobile that they will be including the header/metadata in the base64 string so we'll be able to easily identify the file format. Given this, I'll go ahead and make some changes so that the server is able to accept any file type.

zandre-eng: I've pushed a commit to accommodate uploading any file type ([01141d7](https://github.com/dimagi/connect-id/pull/97/commits/01141d7755748e7a435001ce3edd6d01868af47a)).

users/services.py
calellowitz: > Regarding your thought to have the response be a 500 instead of a 400, I am happy to make this change and can see your point on how this could be considered a server error rather than a client error.

Happy to go forward with this. The error reporting was the big thing. We can always change this later if required.

calellowitz (approved):
Thanks for addressing all the feedback. This looks great!

users/services.py
zandre-eng: Okay cool, I've updated the status code to be 500 now ([ad3aa01](https://github.com/dimagi/connect-id/pull/97/commits/ad3aa011535ba729eeecd02388961c6881093d4b)).

PR #98, opened 2025-05-14
Validate Firebase User ID Token

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1158).
Link to tech spec [here](https://docs.google.com/document/d/1P1J8h1ThL5HS68Q_Dqnsg36g7ReIRUA068wB42Sou_w/edit?tab=t.0).

Mobile will be using a new OTP system with Firebase to generate OTPs for users. As part of those changes, this PR adds a new server endpoint which will be used to validate the unique user token from mobile after a user has entered and verified their OTP. This endpoint can be used to validate the user's phone number by ensuring that the given token is legitimate.

This new POST endpoint takes only a single `id_token` argument which is the generated unique user ID token from Firebase after verifying their OTP. The endpoint will respond with either a 400 and a relevant status code if there is a failure, or a simple 200 HTTP response if the validation succeeded.

Additionally, if the validation succeeded then all invalid configuration sessions for the user's phone number will be removed. This is to ensure that a user can only have one active session after they have verified their number.

As part of testing I created a personal Firebase project and validated the code logic by passing in both a legitimate and fake token from Firebase to ensure that the correct responses get returned.


users/views.py
```diff
@@ -67,6 +68,26 @@ def validate_phone(request):
...
+    user = request.user
+    id_token = request.data.get("id_token")
+    if not id_token:
+        return JsonResponse({"error": ErrorCodes.MISSING_TOKEN}, status=400)
+    try:
+        decoded_token = auth.verify_id_token(id_token)
+    except Exception:
+        return JsonResponse({"error": ErrorCodes.FAILED_VALIDATING_TOKEN}, status=400)
+
+    if not decoded_token.get("uid"):
```
Charl1996, line 82: Why do we need to check this?

users/views.py
```diff
@@ -67,6 +68,26 @@ def validate_phone(request):
     return PhoneDevice.send_otp_httpresponse(phone_number=user.phone_number, user=user)
 
 
+@api_view(["POST"])
```
calellowitz, line 71: This will enforce basic auth on the view. That might be fine for now since we are still lining up the auth work work, but good to keep in mind we will need to update this.

users/views.py
```diff
@@ -67,6 +68,26 @@ def validate_phone(request):
...
+def validate_firebase_id_token(request):
+    user = request.user
```
calellowitz, line 73: Similar to the above, but noting the request won't have a user yet (since it isn't created yet) in it's final form.

users/views.py
```diff
@@ -67,6 +68,26 @@ def validate_phone(request):
...
+    try:
+        decoded_token = auth.verify_id_token(id_token)
+    except Exception:
+        return JsonResponse({"error": ErrorCodes.FAILED_VALIDATING_TOKEN}, status=400)
+
+    if not decoded_token.get("uid"):
+        return JsonResponse({"error": ErrorCodes.INVALID_TOKEN}, status=400)
+    if decoded_token.get("phone_number") != user.phone_number.as_e164:
+        return JsonResponse({"error": ErrorCodes.PHONE_MISMATCH}, status=400)
+    user.phone_validated = True
```
calellowitz, line 86: since we won't have the user at this point in the flow, this will need to work differently

calellowitz (commented):
This generally looks fine, but I think we ran into a bit of a sequencing issue, where the final form of this will be dependent on the auth work (and the rest of the workflow changes that go along with that). I think fine to merge now if we want to do testing of the firebase api part, or leave it for now, and come back once the auth and workflow pieces are done.

users/views.py
zandre-eng: Thanks for pointing that out. I think it's good to leave as is for now. I can always just replace this when the auth PR I'm working on merges in. Otherwise if that merges in first, I'll simply update the decorator in this PR.

users/views.py
zandre-eng: That is a good point. Based on [this comment](https://github.com/dimagi/connect-id/pull/98#discussion_r2099900469) below, I'll remove this and we'll simply stick to doing a token check in this function.

users/views.py
zandre-eng: This was added as a sanity check to make sure that nothing went wrong in the request and that Firebase was able to successfully validate the token with a user. This is not strictly necessary, as I believe `verify_id_token` should throw an exception in most cases if something went wrong, but this was just for that _extra_ peace of mind. 

I don't have strong feelings over this check, so happy to remove this if you think it's just bloating the function unecessarily.

users/views.py
calellowitz: I think we will still want to note that the number has been validated and/or that they passed this step (just likely in the session object rather than the user) , so that we know they are allowed to call the next step, and when we create the user we know to mark the number validated

zandre-eng: Yeah it makes sense to put this in the session object. In the PR for adding the token auth I've [ pushed a commit](https://github.com/dimagi/connect-id/pull/100/commits/f29d792681aa60afbf0a7b0ae8c445bde525d97d) to store this flag. So I'll wait for that PR to get merged first and then do some refactoring here so that we store the phone validated flag on the session object.

zandre-eng: The PR has been merged in and so the validated flag is now being set on the `ConfigurationSession` instance in ([72ef552](https://github.com/dimagi/connect-id/pull/98/commits/72ef5527aff285975236f38cfb28e26f0ea293d5)).

users/views.py
```diff
@@ -70,6 +72,32 @@ def validate_phone(request):
...
+    if not decoded_token.get("uid"):
+        return JsonResponse({"error": ErrorCodes.INVALID_TOKEN}, status=400)
+    if decoded_token.get("phone_number") != request.auth.phone_number.as_e164:
+        return JsonResponse({"error": ErrorCodes.PHONE_MISMATCH}, status=400)
+    request.auth.is_phone_validated = True
+    request.auth.save()
+
+    invalid_sessions = ConfigurationSession.objects.filter(phone_number=request.auth.phone_number).exclude(
+        key=request.auth.key
+    )
```
Charl1996, line 95: I think you can do this in one query:
```
ConfigurationSession.objects.filter(phone_number=request.auth.phone_number).exclude(
        key=request.auth.key
    ).delete()
```

Charl1996 (approved):

PR #102, opened 2025-05-26
Specify account info instead of file path

Charl1996:
I've changed the way we fetch the google credentials by specifying it now directly instead of us giving the package a file path. The reason for this change is that it's more straight forward to handle the credentials in the env file.


Charl1996:
@zandre-eng I've updated the settings.py again to pull in individual pieces of the credentials instead of a dict, which the .env does not support well. It's also more in line with how we populate other credentials like the FCM credentials.

zandre-eng (approved):
Looks good

PR #108, opened 2025-05-28
Verify backup code endpoint

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-1222)

This PR implements the "verify backup code" endpoint (see [here](https://docs.google.com/spreadsheets/d/1J0tiUK0Hf7LSCxne-4wwQHOykIsl_Hhaq95B6LtwRec/edit?gid=1351908939#gid=1351908939)).

A field is added to the `ConfigurationSession` model, `backup_code_attempts`, which tracks how many times a user has attempted the backup code input. The user has 3 attempts (number chosen arbitrarily) before the user's account is orphaned (deactivated).


Charl1996:
One thing to note maybe - this PR changes the current `users/recovery/confirm` view, but maybe this should rather be a new view? Reason being, once we deploy this current users won't be able to recover their accounts. Maybe this is OK since we plan to roll all these new changes out in the next week or two, but is that what we want? Maybe we can temporarily create a new endpoint sitting at `users/backup-code/confirm` (+1 for jargon update) and deprecate the other endpoint after mobile has been released?

Any thoughts?

Charl1996:
I spoke to Parth and we decided it's safer to have a new endpoint, `users/recover/confirm_backup_code`.

users/models.py
```diff
@@ -190,6 +190,10 @@ class ConfigurationSession(models.Model):
     expires = models.DateTimeField()
     phone_number = PhoneNumberField()
     is_phone_validated = models.BooleanField(default=False)
+    backup_code_attempts = models.IntegerField(default=0)
```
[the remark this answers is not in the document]
zandre-eng: I feel like the previous naming was more accurate regarding what this field represents.

Since this is essentially tracking the number of failed attempts a user has tried entering their backup code, this name could be slightly extended to be `failed_backup_code_attempts`.

users/views.py
```diff
@@ -419,23 +419,37 @@ def user_payment_profile(user):
...
 @authentication_classes([SessionTokenAuthentication])
 def confirm_recovery_pin(request):
+    session = request.auth
+
+    if not session.is_phone_validated:
+        return JsonResponse({"error_code": ErrorCodes.PHONE_NOT_VALIDATED}, status=403)
+
     data = request.data
-    user = ConnectUser.objects.get(phone_number=request.auth.phone_number, is_active=True)
+    user = ConnectUser.objects.get(phone_number=session.phone_number, is_active=True)
```
zandre-eng, line 453: Just confirming the worfklow here. Does this endpoint get called before or after `complete_profile`? We'll only be creating the `ConnectUser` instance on `complete_profile`, so the user wouldn't exist yet at this point for a new account configuration. Perhaps we can save the recovery pin on the session itself and then use that for the check here?

Alternatively, we might need to then create the user instance earlier in the workflow so that it is available at this point.

zandre-eng (commented):
Overall the code looks good, just a question regarding the `ConnectUser` instance which gets used in  the view.

users/views.py
```diff
@@ -441,6 +441,43 @@ def confirm_recovery_pin(request):
...
+    session = request.auth
+
+    if not session.is_phone_validated:
+        return JsonResponse({"error_code": ErrorCodes.PHONE_NOT_VALIDATED}, status=403)
+
+    data = request.data
+    user = ConnectUser.objects.get(phone_number=session.phone_number, is_active=True)
+
+    try:
+        if not user.check_recovery_pin(data.get("recovery_pin")):
```
Charl1996, line 456: The `pin` has been updated to `recovery_pin` (see [here](https://docs.google.com/spreadsheets/d/1J0tiUK0Hf7LSCxne-4wwQHOykIsl_Hhaq95B6LtwRec/edit?gid=1351908939#gid=1351908939))

PR #112, opened 2025-05-29
Fix Failing Tests

zandre-eng:
This PR fixes two test cases that fail in the `TestCompleteProfileView` class. This failure was introduced in [this PR](https://github.com/dimagi/connect-id/pull/108) where the `phone_number` and `is_phone_validated` fields were being set for the `valid_token` fixture. This resulted in the following two failures:

- `test_phone_not_validated` - The `is_phone_validated` was not explicitely being set to `False` here
- `test_success` - A phone number was being passed in the session token which was used to create a new user that already exists and is active.

### DIFF
--- a/users/tests/tests.py
+++ b/users/tests/tests.py
@@ -1028,9 +1028,9 @@ class TestCompleteProfileView:
 
     @patch("users.views.upload_photo_to_s3")
     def test_success(self, mock_upload_photo, authed_client_token, valid_token):
-        mock_upload_photo.return_value = None
-        valid_token.is_phone_validated = True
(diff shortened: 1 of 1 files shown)


Charl1996 (approved):
Interesting that I didn't pick this up.

zandre-eng:
@Charl1996 I think what happened was that your PR got merged to `main` and then mine followed afterwards, with the big problem being that your changes weren't merged into mine before I merged it in. That would have shown the failing tests before making it to `main`.

Definitely a good reminder on the value of doing a merge with the latest from `main` before merging in a PR.

PR #124, opened 2025-06-04
Save GPS Data

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1082).

This is the first PR in the effort to add location checking to the `start_configuration` endpoint. The end goal is to be able to compare the user's phone number location with their device location to ensure that both match. This PR starts that off by simply retrieving and saving the user's device location in the session.

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -191,7 +191,7 @@ class ConfigurationSession(models.Model):
     phone_number = PhoneNumberField()
     is_phone_validated = models.BooleanField(default=False)
     failed_backup_code_attempts = models.IntegerField(default=0)
-    gps_location = models.CharField(max_length=100, blank=True, null=True)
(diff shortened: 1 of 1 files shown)


users/models.py
```diff
@@ -191,6 +191,7 @@ class ConfigurationSession(models.Model):
     phone_number = PhoneNumberField()
     is_phone_validated = models.BooleanField(default=False)
     failed_backup_code_attempts = models.IntegerField(default=0)
+    gps_location = models.CharField(max_length=100, blank=True, null=True)
```
pxwxnvermx, line 194: nit: can we add a comment here to document the structure for this location string?

pxwxnvermx (approved):
just a nit, otherwise looks good.

PR #127, opened 2025-06-06
Update 400 -> 403

Charl1996:
We need to return a better status code when the integrity check fails. Also see [this comment](https://dimagi.atlassian.net/browse/QA-7825?focusedCommentId=382685).

### DIFF
--- a/utils/app_integrity/decorators.py
+++ b/utils/app_integrity/decorators.py
@@ -1,4 +1,4 @@
-from django.http import HttpResponseForbidden, JsonResponse
+from django.http import HttpResponseBadRequest, HttpResponseForbidden, JsonResponse
 
 from users.const import TEST_NUMBER_PREFIX
 from utils.app_integrity.const import INTEGRITY_REQUEST_HASH_KEY, INTEGRITY_TOKEN_HEADER_KEY, ErrorCodes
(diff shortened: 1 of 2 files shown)


utils/app_integrity/decorators.py
```diff
@@ -21,7 +21,9 @@ def wrapper(request, *args, **kwargs):
         request_hash = request.headers.get(INTEGRITY_REQUEST_HASH_KEY)
 
         if not (integrity_token and request_hash):
-            return JsonResponse({"error_code": ErrorCodes.INTEGRITY_DATA_MISSING}, status=400)
+            return JsonResponse(
+                {"error_code": ErrorCodes.INTEGRITY_DATA_MISSING}, status=HttpResponseForbidden.status_code
```
zandre-eng, line 25: Curious why this was changed to be a 403 instead of a 400? Looking at the error code it essentially seems like an issue with missing data in the request rather than the user's credentials not being authorised to make the request.

zandre-eng (approved):
Just a comment about one of the changes, but otherwise looks good.

utils/app_integrity/decorators.py
Charl1996: Oh, yeah, you're right! That one should be 400.

zandre-eng (approved):

PR #129, opened 2025-06-12
Don't enforce gps data on start_configuration

Charl1996:
It seems like mobile does not send up GPS data yet, so we shouldn't expect it yet in the payload on `start_configuration`.


Charl1996:
@shubham1g5 made a good point - we should use the api versioning system for any blocking operations like enforcing the GPS data in the payload.

PR #130, opened 2025-06-20
Show integrity error reason in logs

Charl1996:
This is a stab at better logging around the app integrity check to more easily see the reason for failure. 

Would it make sense to add a slug or something to the log as well? We could use the slug to filter more easily, but then we could use the reason string as well. Open to ideas.

### DIFF
--- a/utils/app_integrity/decorators.py
+++ b/utils/app_integrity/decorators.py
@@ -25,7 +25,7 @@ def require_app_integrity(view):
         request_hash = request.headers.get(INTEGRITY_REQUEST_HASH_KEY)
         phone_number = request.data.get("phone_number", "")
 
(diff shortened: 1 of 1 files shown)


calellowitz (commented):
Logging part looks fine, I think the codeql concern is fairly reasonable. You could convince me the number is necessary if we want these to track down individual users, but I am wondering if we could use a session id if we need it to be identifiable or leave it out entirely if we just want to get counts and trends.

zandre-eng (approved):
Nice!

Charl1996:
Merging this. @calellowitz Let me know if there's any followup you'd like me to do apart from what we've discussed over slack, i.e. to send down the reason for the fail to mobile.

PR #131, opened 2025-06-24
Ce/invited user

calellowitz:
This PR does two things, one in each commit

1) Checks if a phone number has been invited to a connect opportunity, if so allows PIN instead of biometric device security and skips the device integrity testing
2) Uses the same field to automatically resend opportunity invites when the account is fully set up.


calellowitz:
Linked Connect PR https://github.com/dimagi/commcare-connect/pull/669

pxwxnvermx (approved):
LGTM. Thanks

hemant10yadav (approved):

PR #139, opened 2025-06-25
Add subcodes

Charl1996:
See [context here](https://docs.google.com/document/d/1duX2l6Yz0Z6Z6wnABZ8r8m9JneyzAjV1PNJuHWL2Vng/edit?tab=t.0).

We need to start specifying error sub codes, which gives us the ability give more detail around why a certain error happened.

This PR adds sub codes to any integrity error response.


Charl1996:
The code scanning results is hung-up on the phone number in the logs.

zandre-eng (approved):

PR #140, opened 2025-06-25
Management Command to Unlock and Generate Backup Code for User

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1380).

This PR adds in a new management command which does the following:
- Given a phone number or user ID, unlocks that user (i.e. marks `is_locked` as false and `is_active` as true).
- Generates a new 6-digit backup code for the unlocked user.
- Optionally marks the currently active user as inactive. This is needed in cases where the user has failed recovery and ended up creating a new account.

### DIFF
--- a/users/management/commands/unlock_and_generate_backup_code.py
+++ b/users/management/commands/unlock_and_generate_backup_code.py
@@ -1,0 +1,68 @@
+import secrets
+
+from django.core.management.base import BaseCommand
+
+from users.models import ConnectUser
+
+
+class Command(BaseCommand):
+    help = "Unlock a user and generate a new backup code for them"
+
+    def add_arguments(self, parser):
+        parser.add_argument("--phone-number", type=str, required=False)
+        parser.add_argument("--inactive-user-id", type=str, required=False, help="User ID of inactive user to unlock")
+        parser.add_argument("--disable-current-active-user", action="store_true")
+
+    def handle(self, *args, **options):
+        phone_number = options.get("phone-number")
+        inactive_user_id = options.get("inactive-user-id")
+        if not phone_number and not inactive_user_id:
+            print("Please provide either a phone number or an inactive user ID")
+            return
+
+        disable_current_active_user = options.get("disable-current-active-user", True)
+
+        inactive_user = get_inactive_user(phone_number, inactive_user_id)
+        user = unlock_user(inactive_user, disable_current_active_user)
+        if user:
+            backup_code = generate_backup_code(user)
+            print(f"User {phone_number} has been unlocked and a backup code has been generated: {backup_code}")
+        else:
+            print(f"Failed to unlock user with phone number {phone_number}.")
+
+
+def get_inactive_user(phone_number, inactive_user_id=None):
+    if inactive_user_id:
+        return ConnectUser.objects.get(id=inactive_user_id)
+
+    # Old accounts might only be set as inactive without being locked, so default to checking for only inactive
+    # if no locked user is found
+    try:
+        inactive_user = ConnectUser.objects.get(phone_number=phone_number, is_active=False, is_locked=True)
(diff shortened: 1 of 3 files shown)


users/management/commands/unlock_and_generate_backup_code.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument("--phone-number", type=str, required=False)
+        parser.add_argument("--inactive-user-id", type=str, required=False, help="User ID of inactive user to unlock")
+        parser.add_argument("--disable-current-active-user", action="store_true")
+
+    def handle(self, *args, **options):
+        phone_number = options.get("phone-number")
+        inactive_user_id = options.get("inactive-user-id")
+        if not phone_number and not inactive_user_id:
```
calellowitz, line 19: It would be good to error if both are provided (or at least if both are provided and they conflict)

users/management/commands/unlock_and_generate_backup_code.py
```diff
@@ -0,0 +1,68 @@
...
+            print("Please provide either a phone number or an inactive user ID")
+            return
```
calellowitz, line 21: not blocking, but this should probably raise an exception or at least return an non-success status code

users/management/commands/unlock_and_generate_backup_code.py
```diff
@@ -0,0 +1,68 @@
...
+
+def get_inactive_user(phone_number, inactive_user_id=None):
+    if inactive_user_id:
+        return ConnectUser.objects.get(id=inactive_user_id)
+
+    # Old accounts might only be set as inactive without being locked, so default to checking for only inactive
+    # if no locked user is found
+    try:
+        inactive_user = ConnectUser.objects.get(phone_number=phone_number, is_active=False, is_locked=True)
+    except (ConnectUser.MultipleObjectsReturned, ConnectUser.DoesNotExist):
```
calellowitz, line 42: This doesn't handle the multiple returned case. It will just return multiple again and reraise the error you caught. I think you will need to return the most recent one in that case (or just demand that they add a user id in that case).

users/management/commands/unlock_and_generate_backup_code.py
```diff
@@ -0,0 +1,68 @@
...
+        inactive_user = ConnectUser.objects.get(phone_number=phone_number, is_active=False)
+    return inactive_user
+
+
+def unlock_user(inactive_user, disable_current_active_user=True):
+    if not inactive_user:
+        return
+
+    if disable_current_active_user:
+        active_user = ConnectUser.objects.filter(phone_number=inactive_user.phone_number, is_active=True).first()
```
calellowitz, line 52: I think this wants to be `.get` since we only expect 1. If it does return more than one active users (which I wouldn't think is possible), we would want to deactivate all of them, not just one of them.

users/management/commands/unlock_and_generate_backup_code.py
```diff
@@ -0,0 +1,68 @@
...
+        inactive_user_id = options.get("inactive-user-id")
+        if not phone_number and not inactive_user_id:
+            print("Please provide either a phone number or an inactive user ID")
+            return
+
+        disable_current_active_user = options.get("disable-current-active-user", True)
+
+        inactive_user = get_inactive_user(phone_number, inactive_user_id)
+        user = unlock_user(inactive_user, disable_current_active_user)
+        if user:
```
calellowitz, line 27: When is user here null? I am a bit confused about some of the error handling. It is only falsy if you pass in a falsy value to the function, in which case it seems like you could check for that before calling unlock. But `get_inactive_user` also can't return a null/falsy value so this seems like an impossible condition.

calellowitz (approved):
None of this is blocking if you need to run the command tomorrow, but it would be great to clean this up in a followup PR.

PR #148, opened 2025-07-01
personal-id twilio sms

sravfeyn:
This implements twilio fallback spec described [here](https://docs.google.com/document/d/1oYsxGDfr9_eRMsEbz-BfBinFzzMbTQK2zgG6afuNHfw/edit?tab=t.0)
Ticket https://dimagi.atlassian.net/browse/CCCT-1420


users/urls.py
```diff
@@ -11,6 +11,8 @@
     path("validate_phone", views.validate_phone, name="validate_phone"),
     path("validate_firebase_id_token", views.validate_firebase_id_token, name="validate_firebase_id_token"),
     path("confirm_otp", views.confirm_otp, name="confirm_otp"),
+    path("send_otp_personal_id", views.send_otp_personal_id, name="send_otp_personal_id"),
+    path("confirm_otp_personal_id", views.confirm_otp_personal_id, name="confirm_otp_personal_id"),
```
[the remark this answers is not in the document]
calellowitz: As always with these, it would be good to coordinate these URLs with the mobile team. I know these depart from the spec and otherwise seem fine, but the key is making sure mobile is aware of, and ok with, the changes. I don't have any specific issues with the options chosen (I see Zandre's point, but also Shubham's previous desire to remove twilio branding since these may not be twilio forever), but good to make sure mobile is ok with them as well.

users/views.py
```diff
@@ -833,3 +834,24 @@ def check_user_similarity(request):
...
+
+
+@api_view(["POST"])
+@authentication_classes([SessionTokenAuthentication])
+def confirm_otp_personal_id(request):
+    device = SessionPhoneDevice.objects.get(phone_number=request.auth.phone_number, user_session=request.auth)
+    data = request.data
+    verified = device.verify_token(data.get("otp"))
+    if not verified:
+        return JsonResponse({"error": "OTP token is incorrect"}, status=401)
```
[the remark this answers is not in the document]
calellowitz: agree that consistency in error messaging sounds like the right path

users/models.py
```diff
@@ -238,6 +246,16 @@ def country_code(self):
         return address.get("country_code")
 
 
+class SessionPhoneDevice(BasePhoneDevice):
+    user_session = models.ForeignKey(ConfigurationSession, on_delete=models.CASCADE)
```
calellowitz, line 250: nit: `user_session` is not quite accurate here. `session` or `configuration_session` seem more correct.

users/models.py
```diff
@@ -238,6 +246,16 @@ def country_code(self):
...
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, null=True, blank=True)
```
calellowitz, line 251: why add the nullable user model here?

users/views.py
```diff
@@ -833,3 +834,24 @@ def check_user_similarity(request):
             "photo": existing_user.get_photo() if is_same_user else "",
         }
     )
+
+
+@api_view(["POST"])
+@authentication_classes([SessionTokenAuthentication])
+def send_otp_personal_id(request):
+    otp_device, _ = SessionPhoneDevice.objects.get_or_create(phone_number=request.auth.phone_number, user_session=request.auth)
```
calellowitz, line 842: can this key only off `request.auth`? The phone number can't change for an auth session so it seems like the first field is redundant

users/models.py
```diff
@@ -238,6 +246,16 @@ def country_code(self):
...
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["phone_number", "user_session"], name="phone_number_user_session")
```
calellowitz, line 255: I think this should only needs to be unique session rather than the combo of session and phone number

users/views.py
```diff
@@ -833,3 +834,24 @@ def check_user_similarity(request):
...
+    otp_device.generate_challenge()
+    return HttpResponse()
+
+
+@api_view(["POST"])
+@authentication_classes([SessionTokenAuthentication])
+def confirm_otp_personal_id(request):
+    device = SessionPhoneDevice.objects.get(phone_number=request.auth.phone_number, user_session=request.auth)
```
calellowitz, line 850: same question about only pass the session

calellowitz (approved):
Left a few suggestions, nothing blocking necessarily but I think they are all improvements.

users/views.py
sravfeyn: Similar to PhoneDevice, this requires phone-number to generate the OTP, so we do need to pass the phone-number.

users/views.py
sravfeyn: Responded [above](https://github.com/dimagi/connect-id/pull/148#discussion_r2179218321)

users/urls.py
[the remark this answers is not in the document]
sravfeyn: Yeah, the naming has been up for discussion [here](https://docs.google.com/document/d/1oYsxGDfr9_eRMsEbz-BfBinFzzMbTQK2zgG6afuNHfw/edit?disco=AAABm1D3OB8) on the doc. Definitely will need to align with mobile on the final endpoints, so that they can use the same.

May be `_sms` suffix could be an option we could align on?

users/tests/tests.py
```diff
@@ -1170,3 +1170,94 @@ def test_existing_account_deactivation(self, mock_upload_photo, authed_client_to
...
+
+        with patch("users.models.SessionPhoneDevice.generate_challenge") as mock_generate_challenge:
+            response = authed_client_token.post(self.url)
+            assert response.status_code == 200
+            assert SessionPhoneDevice.objects.filter(user_session=valid_token).exists()
+            mock_generate_challenge.assert_called_once()
+
+
+@pytest.mark.django_db
+class TestConfirmOtpPersonalId:
```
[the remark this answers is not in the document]
sravfeyn: This would be a redundant test (since it's handled by SessionTokenAuthentication) which is tested [here](https://github.com/dimagi/connect-id/blob/main/users/tests/tests.py#L803)

users/models.py
sravfeyn: Done

users/views.py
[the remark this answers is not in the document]
sravfeyn: This was actually made consistent with the existing `confirm_otp` which does send this message instead of the error code. 

I have updated to send error code to keep consistent with `validate_firebase_id_token` instead

users/tests/tests.py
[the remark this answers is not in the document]
zandre-eng: Sorry, what I meant by token was the OTP token that gets generated (not the session token). I see it also has an expiry date.

sravfeyn:
@calellowitz @zandre-eng Thanks for the review, updated per your comments.

zandre-eng (approved):

users/models.py
calellowitz: Since we will never use this field, we can remove it from the model by setting it to `None`

users/models.py
calellowitz: I don't think so, regardless of the answer to the get or create ones. I think we still want to enforce that we only create one of these per session (likely with a `OneToOneField`

users/views.py
calellowitz: Even if we want to pass a phone number, I don't think it should be part of the `get_or_create` block because if a session is passed with a different phone number we don't want a new device to be created. That said, I don't think the phone should be necessary. I think we can remove the field and replace it with a property for this model, but if we can't we can still remove it from the `get_or_create`. 

> This is modeled exactly as PhoneDevice which could also take just the user object.

I don't think this is quite the same. We pass the phone number with the user because users could have multiple phone devices (one for the primary phone and one for the secondary phone), in this case there can only be one.

users/views.py
calellowitz: Same point as above. I think the `get` should only need the session because the session should be unique, regardless of whether the phone number exists.

calellowitz:
The comments still aren't blocking, but I do still think they would be an improvement to this code

PR #149, opened 2025-07-02
Update Credential Model

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1371).
Link to tech spec [here](https://docs.google.com/document/d/1lHHZEx7_MOWxu1RxH1wXb7wGuLWHHp68wWx-jtWwBX8/edit?tab=t.0).

This PR updates the `Credential` model in preparation for having the ability to receive user work history credentials from CommCare HQ.


users/views.py
```diff
@@ -643,7 +643,11 @@ def post(self, request, *args, **kwargs):
         credential_name = request.data["credential"]
         slug = f"{credential_name.lower().replace(' ', '_')}_{org_slug}"
         credential, _ = Credential.objects.get_or_create(
-            name=credential_name, organization_slug=org_slug, defaults={"slug": slug}
+            name=credential_name,
+            issuing_authority_slug=org_slug,
+            defaults={"slug": slug},
+            issuing_authority_type=Credential.IssuingAuthorityType.CONNECT,
+            type=Credential.CredentialType.LEARN,
```
zandre-eng, line 650: From what I understand, currently these will just be Learn credentials from Connect, and so I've put it as such here. Noting that this endpoint will still be updated in a follow-up ticket to allow for HQ credentials as well.

users/models.py
```diff
@@ -172,9 +172,22 @@ class RecoverySteps(models.TextChoices):
...
+        DELIVER = "deliver", "deliver"
+
     name = models.CharField(max_length=300)
     slug = models.CharField(max_length=100)
-    organization_slug = models.CharField(max_length=255)
+    issuing_authority_slug = models.CharField(max_length=255)  # Either app_id for HQ creds, or org slug for CCC creds
+    issuing_authority_type = models.CharField(max_length=25, choices=IssuingAuthorityType.choices)
+    type = models.CharField(max_length=25, choices=CredentialType.choices)
+    created_on = models.DateTimeField(auto_now_add=True)
+    level = models.CharField(max_length=50, null=True, blank=True)  # Credential level/code (e.g. 3_MONTHS_ACTIVE)
```
zandre-eng, line 190: Going through the tech spec, I was a little confused as to what we ultimately decided on adding to the model. Specifically, I'm still a little uncertain about where the `type` credential comes from.

@Charl1996 could you take a look and let me know if the fields I've added look right according to what you imagined in the spec, or if there's something I'm missing.

users/models.py
```diff
@@ -172,9 +172,22 @@ class RecoverySteps(models.TextChoices):
...
+
+    class CredentialType(models.TextChoices):
+        APP_ACTIVITY = "app_activity", "app_activity"
+        LEARN = "learn", "learn"
+        DELIVER = "deliver", "deliver"
+
     name = models.CharField(max_length=300)
     slug = models.CharField(max_length=100)
-    organization_slug = models.CharField(max_length=255)
+    issuing_authority_slug = models.CharField(max_length=255)  # Either app_id for HQ creds, or org slug for CCC creds
```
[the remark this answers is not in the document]
zandre-eng: Do you mean we keep it as `organization_slug` or we remove this field entirely? If the latter, where do we then save the `app_id` or `org_slug`?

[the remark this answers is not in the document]
Charl1996: Or are you repurposing that field for app_id or org_slug?

users/views.py
Charl1996: > From what I understand, currently these will just be Learn credentials from Connect

I don't think the current credentials are "learn" credentials. As far as I know the current credentials are Connect organizations that have created a "credential" on Connect (which is eventually stored in PersonalID) that is issued to the users that served an opportunity for that organization. The idea was so they can go "I want to invite all users who have credential X for this new opportunity".

In terms of the learn credentials, I think that's stored currently on Connect (but now we want to store them on PersonalID).

That said, I'm not sure what to set the default to though. Maybe "deliver" is closer to reality.

users/models.py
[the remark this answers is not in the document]
zandre-eng: > Or are you repurposing that field to store the id of "the thing that the issuing authority cares about"?

Yeah essentially. So for HQ credentials it would be the `app_id` and for Connect credentials it can be the `org_id` or `org_slug`.

users/models.py
```diff
@@ -172,9 +172,34 @@ class RecoverySteps(models.TextChoices):
...
+        LEARN = "LEARN", "LEARN"
+        DELIVER = "DELIVER", "DELIVER"
+
+    uuid = models.UUIDField(default=uuid4)
+    title = models.CharField(max_length=300)
+    issuing_authority = models.CharField(max_length=50, choices=IssuingAuthorityTypes.choices)
+    created_at = models.DateTimeField(auto_now_add=True)
+    level = models.CharField(max_length=50, blank=True, null=True)  # credential level/code (e.g. 3_MONTHS_ACTIVE)
+    type = models.CharField(max_length=50, choices=CredentialTypes.choices)
+    app_or_opp_id = models.CharField(max_length=50, blank=True, null=True)
```
Charl1996, line 190: I'm almost more in favour of splitting this in two fields now that I see it like this. Or rename this to something like `credential_entity_id`, signifying _"this is the context or thing that the credential is linked to"_ ... or something like that. But if you're not convinced we should change it I can be convinced otherwise.

@calellowitz For context, we thought of storing the "opportunity_id" or "app_id" in this one field encapsulating these two concepts, because we can simply derive whether it's an `app_id` or `opp_id` based on the `issuing_authority` when compiling the "list credentials" response for mobile.

calellowitz: I am with charl that I don't love the name. I think if we are going to combine these fields, we are likely to have future types with different values in this field so having a name that encapsulates all possible values makes sense, but it feels in the short term like splitting it out may be simpler, especially if we want to ever include both (like a learn credential that shows both the learn app and the opportunity). basically, anything with "or" feels like we are just digging a hole that the next time we add a possibility we will have to rename the column

zandre-eng: Thanks for your inputs @Charl1996 and @calellowitz. The reasoning makes sense to me. I've gone ahead and split out this field to be two separate fields ([4b59b3c](https://github.com/dimagi/connect-id/pull/149/commits/4b59b3c2c3739f0d0e4b4efab4e887f9ec027be1)).

calellowitz (commented):
What happens to existing rows in this table when the migraiton is run? It seems like they will be incompatible with the new schema

calellowitz:
Given that that will delete any useful info to identify the old credential, I think it is worth explicitly deleting them. Otherwise we will have a bunch of misleading data in the table, that will look like real credentials but won't correspond to the data inside them. Then we can also remove the defaults from `issuing_authority` and `type` since we probably don't want default values there (we will want those to be explicitly set when a credential is created).

users/models.py
```diff
@@ -178,9 +178,23 @@ class RecoverySteps(models.TextChoices):
...
+    class CredentialTypes(models.TextChoices):
+        APP_ACTIVITY = "APP_ACTIVITY", "APP_ACTIVITY"
+        LEARN = "LEARN", "LEARN"
+        DELIVER = "DELIVER", "DELIVER"
+
+    uuid = models.UUIDField(default=uuid4)
+    title = models.CharField(max_length=300)
+    issuing_authority = models.CharField(max_length=50, choices=IssuingAuthorityTypes.choices)
+    created_at = models.DateTimeField(auto_now_add=True)
+    level = models.CharField(max_length=50, blank=True, null=True)  # credential level/code (e.g. 3_MONTHS_ACTIVE)
```
calellowitz, line 194: Why is this nullable?

users/models.py
```diff
@@ -178,9 +178,23 @@ class RecoverySteps(models.TextChoices):
...
+        CONNECT = "CONNECT", "CONNECT"
+        HQ = "HQ", "HQ"
+
+    class CredentialTypes(models.TextChoices):
+        APP_ACTIVITY = "APP_ACTIVITY", "APP_ACTIVITY"
+        LEARN = "LEARN", "LEARN"
+        DELIVER = "DELIVER", "DELIVER"
+
+    uuid = models.UUIDField(default=uuid4)
+    title = models.CharField(max_length=300)
```
calellowitz, line 191: what is title? Is this just a name? If so, any reason to rename?

users/models.py
```diff
@@ -178,9 +178,23 @@ class RecoverySteps(models.TextChoices):
 
 
 class Credential(models.Model):
```
calellowitz, line 180: I think we want a unique constraint on whatever set of fields uniquely defines a credential. I suspect that is (authority, level, type, <identifier>) where identifier is a new column that allows the issuer to uniquely describe the credential. It is probably the app id for hq issued credentials but could be other ids in other places. The tech spec calls this `slug` which is also a reasonable name

users/views.py
```diff
@@ -749,12 +749,12 @@ class FetchCredentials(ClientProtectedResourceMixin, View):
     required_scopes = ["user_fetch"]
 
     def get(self, request):
```
calellowitz, line 751: I don't think these changes are helpful. They won't work without changes in connect, and pretty meaningfully change what the view does. Given that we have removed usage of this view from connect for now, I would recommend leaving this as is or remove it entirely so that it is clear that it has not been updated, rather than misleading a potential future user into thinking it will work as before.

calellowitz (changes requested):
Apologies for so many new comments in a second review. After seeing it in use in the other PR, I noticed a few more things.

users/models.py
zandre-eng: The `slug` referenced in the tech spec will be saved to either the `app_id` or `opportunity_id` field depending on where the credential is coming from. Given this, would it make sense to then put a constraint as (issuing_authority, level, type, app_id, opportunity_id)?

users/views.py
zandre-eng: That makes sense. Given that this view isn't being used, I'm in favour of removing it entirely. Leaving it as is might also be confusing to devs who notice that this view is trying to update fields that don't exist anymore.

zandre-eng:
> Given that that will delete any useful info to identify the old credential, I think it is worth explicitly deleting them. Otherwise we will have a bunch of misleading data in the table, that will look like real credentials but won't correspond to the data inside them. Then we can also remove the defaults from issuing_authority and type since we probably don't want default values there (we will want those to be explicitly set when a credential is created).

@calellowitz That sounds reasonable given that existing credentials aren't being used at all. I'll modify the migration to delete all the existing records from the table before making the schema changes.

users/models.py
zandre-eng: Updated in [46586dd](https://github.com/dimagi/connect-id/pull/149/commits/46586dd6b51a135a8a31efb8d9ecc6071915884e).

users/views.py
calellowitz: thanks, sounds good.

users/models.py
calellowitz: I think of this as equivalent to `title`, we don't call it `app_name` even though that is what it is for these credentials, because we know it could mean other things for other credentials but will still be useful.

users/models.py
```diff
@@ -178,9 +178,28 @@ class RecoverySteps(models.TextChoices):
...
+    class CredentialTypes(models.TextChoices):
+        APP_ACTIVITY = "APP_ACTIVITY", "APP_ACTIVITY"
+        LEARN = "LEARN", "LEARN"
+        DELIVER = "DELIVER", "DELIVER"
+
+    uuid = models.UUIDField(default=uuid4)
+    title = models.CharField(max_length=300)
+    issuing_authority = models.CharField(max_length=50, choices=IssuingAuthorityTypes.choices)
+    created_at = models.DateTimeField(auto_now_add=True)
+    level = models.CharField(max_length=50)  # credential level/code (e.g. 3_MONTHS_ACTIVE)
```
sravfeyn, line 194: Is this a free-text, or are there set constant levels?

users/models.py
```diff
@@ -178,9 +178,28 @@ class RecoverySteps(models.TextChoices):
...
+    type = models.CharField(max_length=50, choices=CredentialTypes.choices)
+    app_id = models.CharField(max_length=50, blank=True, null=True)
+    opportunity_id = models.CharField(max_length=50, blank=True, null=True)
+    slug = models.CharField(max_length=50)
+    issuer_environment = models.CharField(max_length=50)  # e.g. "production", "staging"
```
sravfeyn, line 199: Shouldn't this be a choice with those constants?

users/models.py
```diff
@@ -178,9 +178,28 @@ class RecoverySteps(models.TextChoices):
...
+
```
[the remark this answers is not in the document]
Charl1996: `UserCredential` links the `ConnectUser` to the `Credential`, so I don't think adding `phone_number` here is necessary.

Charl1996:
Looks good from my side. Will wait for [this update](https://github.com/dimagi/connect-id/pull/149#discussion_r2204031657).

users/models.py
zandre-eng: I believe we will have some set constant levels (e.g. 3_MONTH_ACTIVE, 6_MONTH_ACTIVE), but these will be changing in the future as we add more possible levels for the credential. 

@Charl1996 curious to hear your thoughts here regarding keeping this free-text or making it a choices field?

Charl1996: If we make it a set of choices we will need to keep it up to date. But maybe that's OK, since mobile will have to also know how to intepret whatever text is in there, so updates will always have to be made at least somewhere.

zandre-eng: My only concern with having to keep it up-to-date is that we'll have to make a new migration each time we update this field. Perhaps this isn't too bad, but I feel like it would unnecessarily bloat our migrations directory. 

I'm also not sure whether we know exactly what levels we will support, or what their format looks like yet?

Charl1996: > I feel like it would unnecessarily bloat our migrations directory.

Yeah. I don't have a strong preference for either, so maybe don't bother for now?

zandre-eng: That sounds good to me. Let's leave it as a free text field for now.

zandre-eng:
> Looks good from my side. Will wait for [this update](https://github.com/dimagi/connect-id/pull/149#discussion_r2204031657).

@Charl1996 The comment has been addressed and is ready for a final review.

users/models.py
```diff
@@ -178,9 +178,33 @@ class RecoverySteps(models.TextChoices):
...
 
 class Credential(models.Model):
-    name = models.CharField(max_length=300)
-    slug = models.CharField(max_length=100)
-    organization_slug = models.CharField(max_length=255)
+    class IssuingAuthorityTypes(models.TextChoices):
+        CONNECT = "CONNECT", "CONNECT"
+        HQ = "HQ", "HQ"
+
+    class IssuingAuthorityEnvironments(models.TextChoices):
```
calellowitz, line 185: Is the idea that these will be used for both HQ and connect? that prod means prod connect or prod hq? and similar for staging?

users/models.py
```diff
@@ -178,9 +178,33 @@ class RecoverySteps(models.TextChoices):
...
+    uuid = models.UUIDField(default=uuid4)
+    title = models.CharField(max_length=300)
+    issuing_authority = models.CharField(max_length=50, choices=IssuingAuthorityTypes.choices)
+    created_at = models.DateTimeField(auto_now_add=True)
+    level = models.CharField(max_length=50)  # credential level/code (e.g. 3_MONTHS_ACTIVE)
+    type = models.CharField(max_length=50, choices=CredentialTypes.choices)
+    app_id = models.CharField(max_length=50, blank=True, null=True)
+    opportunity_id = models.CharField(max_length=50, blank=True, null=True)
+    slug = models.CharField(max_length=50)
+    issuer_environment = models.CharField(max_length=50, choices=IssuingAuthorityEnvironments.choices)
```
calellowitz, line 204: I was imagining the `issuing_authority` encompassing both of these fields (having the authority be "prod hq" or "www.commcarehq.org", for example). If they are separate, then the unique constraints need to include the environment as well for the same reason we wanted to add the field in the first place. App IDs are not guaranteed to be unique across environemnts so if an app with the same id existed in staging and prod, their credentials will not be distinguishable.

One reason to keep the fields separate is if we thought we wanted to fork behavior somewhere on hq vs connect.

calellowitz (commented):
Left a couple thoughts

users/models.py
zandre-eng: Actually, given [this](https://github.com/dimagi/connect-id/pull/155#discussion_r2212766856) discussion in a related PR it sounds like we can get rid of these two separate fields and can instead rely on using a foreign key instead.

users/models.py
```diff
@@ -177,10 +178,39 @@ class RecoverySteps(models.TextChoices):
...
+        HQ = "HQ", "HQ"
+
+    class IssuingAuthorityEnvironments(models.TextChoices):
+        PRODUCTION = "production", "production"
+        STAGING = "staging", "staging"
+        INDIA = "india", "india"
+
+    issuing_authority = models.CharField(max_length=50, choices=IssuingAuthorityTypes.choices)
+    issuer_environment = models.CharField(max_length=50, choices=IssuingAuthorityEnvironments.choices)
+    oauth_application = models.ForeignKey(Application, on_delete=models.PROTECT)
```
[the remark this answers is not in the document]
calellowitz: I possibly missed this conversation but I think it depends if these are pushed or pulled from HQ (and connect). If those servers are pushing, oauth makes the most sense for authentication, and we will want to add new apps, as you mention (or maybe use existing ones? but I don't think that changes the answer). If we are pulling from those servers, the application might not be relevant, and so could be nullable or non-existent.

calellowitz (approved):
tried to answer the question, but nothing blocking on my end, if hq is pushing

users/models.py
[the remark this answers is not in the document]
zandre-eng: Based on offline discussion I have made some changes so that the foreign key is to `MessageServer`, since that is where we store the credentials from HQ. Additionally, I have made a new auth class to verify that the issuing authority instance related to a client ID exists (as well as validating the client ID and secret). [ecf2869](https://github.com/dimagi/connect-id/pull/149/commits/ecf286958c2e0457dfde85cc4a57bb3315879162)

users/models.py
```diff
@@ -177,10 +177,55 @@ class RecoverySteps(models.TextChoices):
     step = models.TextField(choices=RecoverySteps.choices)
 
 
+class IssuingCredentialsAuth(models.Model):
```
zandre-eng, line 180: I wasn't entirely sure what would be a good name here, so if any folks have objections with this naming I would be happy to change it.

users/models.py
```diff
@@ -178,6 +177,18 @@ class RecoverySteps(models.TextChoices):
...
```
calellowitz, line 180: I think this approach is fine, but isn't quite what we talked about. I thought we discussed having one model (similar to this one, but called `ServerKeys` or similar) that _both_ the `MessageServer` and `IssuingAuthority` would foreign key to, so that we could share credentials for the same physical server between multiple auth types and models.

If we don't want to do that, and want fully separate credentials, then I think this can be part of the `IssuingAuthority` model, since the main advantage of separating it was to allow it to be shared.

calellowitz (approved):
Comments aren't blocking because I think this is a valid approach but I thought we had agreed on something slightly different

users/models.py
```diff
@@ -177,7 +177,7 @@ class RecoverySteps(models.TextChoices):
     step = models.TextField(choices=RecoverySteps.choices)
 
 
-class IssuingCredentialsAuth(models.Model):
+class ServerKeys(models.Model):
     name = models.CharField(max_length=255)
     client_id = models.CharField(max_length=100, unique=True, db_index=True)
```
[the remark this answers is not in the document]
zandre-eng: [c1ed994](https://github.com/dimagi/connect-id/pull/149/commits/c1ed9940beead1498fb654336debf20815a04e99)

users/auth.py
```diff
@@ -28,12 +28,12 @@ def authenticate_credentials(self, key):
         return (user, token)
 
 
-class IssuingCredentialsAuthentication(BasicAuthentication):
+class ServerKeysAuthentication(BasicAuthentication):
```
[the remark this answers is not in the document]
zandre-eng: [4539672](https://github.com/dimagi/connect-id/pull/149/commits/453967240270a17cfdfbec6f3901fbac95f3f36e)

zandre-eng:
@calellowitz I have made changes so that `MessageServer` now uses the `server_credentials` field instead. I have kept the `serv er_id` and `secret_key` fields for now until we are able to move the credentials over to a `ServerKeys` instance.

utils/rest_framework.py
```diff
@@ -30,10 +30,10 @@ class MessagingServerAuth(BasicAuthentication):
 
     def authenticate_credentials(self, userid, password, request=None):
         try:
-            server = MessageServer.objects.get(server_id=userid)
-        except MessageServer.DoesNotExist:
+            server_keys = ServerKeys.objects.get(client_id=userid)
```
calellowitz, line 33: I think this still needs to use `MessageServer` much like the other auth uses `IssuingAuthority` so we know it has been set up for messaging, not just that credentials exist

zandre-eng:
> Is there a migration for this? Otherwise it looks like this will break message auth until that happens

@calellowitz No migration for this currently. That's why I originally opted to leave everything to use the `server_id` and `secret_key` from `MessageServer` until we can manually create new `ServerKeys` instances.

Originally I thought creating a migration to create and link the `server_credentials` might be a bit awkward, but after looking into it some more, I think we're able to implement something that would suffice ([6bd9fa4](https://github.com/dimagi/connect-id/pull/149/commits/6bd9fa457f5ebe5e9acc9b4e8b93ad868af6a3e0)). Let me know what you think.

utils/rest_framework.py
zandre-eng: [744ba67](https://github.com/dimagi/connect-id/pull/149/commits/744ba67f36b35d30c00d379373e3353318376b4f)

zandre-eng:
> Will this go through QA?

@pxwxnvermx I think we would only be able to QA it once the entire feature has been implemented. This is because we'll need the mobile component to be done so that QA has something to test with.

PR #151, opened 2025-07-03
Github deploy action

sravfeyn:
It's been hard to deploy from Mac (with Apple Silicon), this should make it easy to deploy. I have adapted this from commcare-connect github deploy action. Currently, it doesn't have the slackbot integration. PR for that is seperate https://github.com/dimagi/connect-id/pull/152

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,56 @@
+name: Deploy
+
+env:
+  AWS_REGION: 'us-east-1'
+
+permissions:
+  id-token: write
+  contents: read
+
+on:
+  #  workflow_run:
+  #    workflows: [CI]
+  #    types: [completed]
+  #    branches: [main]
+  workflow_dispatch:
+    inputs:
+      environment:
+        required: true
+        default: 'staging'
+        type: choice
+        options:
+          - production
+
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,56 @@
...
+  #    types: [completed]
+  #    branches: [main]
+  workflow_dispatch:
+    inputs:
+      environment:
+        required: true
+        default: 'staging'
+        type: choice
+        options:
+          - production
```
sravfeyn, line 22: This is pretty much redundant and not used, but I have kept it anyway

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  #  workflow_run:
+  #    workflows: [CI]
+  #    types: [completed]
+  #    branches: [main]
+  workflow_dispatch:
+    inputs:
+      environment:
+        required: true
+        default: 'staging'
```
hemant10yadav, line 19: Is it intentional to set the default to `staging` even though `staging` is not one of the available choices?

sravfeyn: Updated

PR #152, opened 2025-07-03
Add slack integration

sravfeyn:
We can merge this once a SLACK_BOT_TOKEN is added to github secrets.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -51,6 +51,85 @@ jobs:
           role-session-name: GithubDeploy
           aws-region: ${{ env.AWS_REGION }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -51,6 +51,85 @@ jobs:
           role-session-name: GithubDeploy
           aws-region: ${{ env.AWS_REGION }}
 
+      - id: slack
+        name: Notify Deploy Start
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          channel-id: 'C05APTXCTSM'
```
Charl1996, line 58: I guess this channel ID isn't secret?

PR #155, opened 2025-07-07
Modify Add Credential Endpoint

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1370?atlOrigin=eyJpIjoiYTUxNDU3OTQxMzhmNGVlNjkzNzIxMmRjNzE3OTI4YmQiLCJwIjoiaiJ9).
Link to tech spec [here](https://docs.google.com/document/d/1lHHZEx7_MOWxu1RxH1wXb7wGuLWHHp68wWx-jtWwBX8/edit?tab=t.0#heading=h.6pi4ff5gli3e).

This PR modifies the existing `add_credential` endpoint as part of the wider changes to support PersonalID Worker History. At a high-level the changes that were made are:
- Support for new fields in the `Credential` model. This includes supporting both `CONNECT` and `HQ` credentials.
- Support for bulk add.

The expected request payload should be in this format:
```
{
    "credentials": [
        {
            "title": "Credential Title",
            "type": "APP_ACTIVITY", // "LEARN" or "DELIVER" for Connect, and "APP_ACTIVITY" for HQ
            "level": "3_MONTH_ACTIVE", // Code/Qualifier for credential
            "issuer": "HQ",  // "HQ" or "CONNECT"
            "app_id": "e2c930de2eaf465b81d083241d870ef9", // or None if issuer is "CONNECT"
            "opp_id": "51f721535c7340dabb2a39a0c2a9709c", // or None if issuer is "HQ"
            "users": ["123456789", "987654321"] // User phone numbers
        },
        ...
    ]
}
```

This PR does significantly change the view, however this is a safe change to make since (1) credentials aren't being used at the moment and (2) adding credentials from Connect has been disabled.


users/views.py
```diff
@@ -648,16 +649,23 @@ class AddCredential(APIView):
...
-        credential_name = request.data["credential"]
-        credential, _ = Credential.objects.get_or_create(
-            title=credential_name,
-            issuing_authority=Credential.IssuingAuthorityTypes.CONNECT,
-            type=Credential.CredentialTypes.DELIVER,
-        )
-        users = ConnectUser.objects.filter(phone_number__in=phone_numbers, is_active=True)
-        for user in users:
-            UserCredential.add_credential(user, credential, request)
+        data = request.data
```
calellowitz, line 652: nit: why assign this to a new variable?

users/views.py
```diff
@@ -648,16 +649,23 @@ class AddCredential(APIView):
...
+        for cred in data.get("credentials", []):
+            try:
+                credential, _ = Credential.objects.get_or_create(
```
calellowitz, line 655: Do all of these fields uniquely identify a credentail or just a smaller set? For example, if the title is changed do we want a new credential created, as will happen here, or to use the old one?

users/views.py
```diff
@@ -648,16 +649,23 @@ class AddCredential(APIView):
...
+                    title=cred.get("title"),
+                    type=cred.get("type"),
+                    level=cred.get("level"),
+                    issuing_authority=cred.get("issuer"),
```
calellowitz, line 659: Can we infer `issuer` from the request? Otherwise HQ could issue credentials claiming to be connect.

calellowitz (commented):
a few questions

users/views.py
zandre-eng: I think I was expecting to use this variable more than I actually did in this function. Given that this is only being used once, it makes sense to remove this variable altogether.

users/views.py
zandre-eng: Based on the discussion in the [model PR](https://github.com/dimagi/connect-id/pull/149#discussion_r2193683105) I think having only type, level, issuing_authority, app_id, and opportunity_id be the unique identifiers sounds reasonable.

With the above being said, we can then remove title here since I don't think changing the name alone should result in an entirely new credential. Though I am curious to hear if @Charl1996 has any thoughts when he is back online.

users/views.py
calellowitz: I was thinking based on the authentication. We should be able to tell which server is talking to us based on the credentials provided (I think we do something similar with messaging if you want to look to those views for an example but you don't have to do it in exactly the same way). Each HQ and connect will use different client credentials for oauth

zandre-eng: Having a look around, I do see there is a `get_current_messaging_server()` which looks like what we want to achieve. Using that for inspiration we could then do something like:
```
    import base64
    from oauth2_provider.models import Application
    from django.shortcuts import get_object_or_404
    auth_header = request.headers.get("authorization")
    encoded_credentials = auth_header.split(" ")[1]
    decoded_credentials = base64.b64decode(encoded_credentials).decode("utf-8")
    client_id, client_secret = decoded_credentials.split(":")
    server = get_object_or_404(Application, client_id=client_id)
```

Does the above look like what you were thinking of? 

It seems like we use `MessageServer` to hold our client ID and secret for HQ, so also just checking in whether we should swap the above `Application` reference for `MessageServer`? If we do, I think we'll also just need to change the authentication class to use `MessagingServerAuth` instead.

users/views.py
zandre-eng: Title removed from query in [92acd28](https://github.com/dimagi/connect-id/pull/155/commits/92acd28ba23820ba1657695162d935ab07993c15).

users/tests/tests.py
```diff
@@ -355,6 +365,99 @@ def test_no_recovery_phone(self, client, user):
         assert json.loads(response.content) == {"error": NO_RECOVERY_PHONE_ERROR}
 
 
+@pytest.mark.django_db
+class TestAddCredential:
+    endpoint = reverse("add_credential")
+
+    def test_no_auth(self, client):
```
sravfeyn, line 372: Might be worth adding a test case for idempotency or weird requests?

users/views.py
```diff
@@ -648,16 +649,23 @@ class AddCredential(APIView):
...
+                credential, _ = Credential.objects.get_or_create(
+                    type=cred.get("type"),
+                    level=cred.get("level"),
+                    issuing_authority=cred.get("issuer"),
+                    app_id=cred.get("app_id"),
+                    opportunity_id=cred.get("opp_id"),
+                )
+                credential.title = cred.get("title")
+                credential.save()
+            except IntegrityError:
```
Charl1996, line 663: Currently it seems that if any one of the credentials is not specified correctly the server will respond without going any further. I think we should probably absorb the error and continue for as long as we can to create the other credentials, then return a list of failed credentials to the caller. 

What do you think?

zandre-eng: That sounds fair, and would at least help unblock the credentials that are specified correctly. 

A question that this does raise however, is what response code should we send back if there were invalid credentials? If we send a 400 back then this could be interpreted as the entire request having failed, but if we send a 200 back then the caller might think everything went fine (unless we always explicitely check the response to see if the list of failed credentials contains any values).

I'm opted to go for a 200 response, but curious to hear what you think?

users/tests/tests.py
zandre-eng: Addressed in [f8c23ab](https://github.com/dimagi/connect-id/pull/155/commits/f8c23ab2128532f2e1f721a855d389dd46ab4315).

users/views.py
zandre-eng: I've pushed a commit with an implementation for inferring the issuer name based on the credentials ([937256b](https://github.com/dimagi/connect-id/pull/155/commits/937256b51b1d696e5db552b1a3ba18bfa6ee7e08)).

@calellowitz @sravfeyn Curious to hear if you have any thoughts on the taken approach?

users/views.py
calellowitz: 200 is generally the correct response in these cases where it partially succeeds and a list of failures is included

users/views.py
```diff
@@ -644,20 +647,48 @@ def get(self, request, *args, **kwargs):
...
+                break
+        for cred in creds:
+            try:
+                credential, _ = Credential.objects.get_or_create(
+                    type=cred.get("type"),
+                    level=cred.get("level"),
+                    issuing_authority=issuer,
+                    slug=cred.get("app_id"),
+                )
+                credential.issuer_environment = cred.get("issuer_environment")
```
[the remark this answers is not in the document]
zandre-eng: Given [this](https://github.com/dimagi/connect-id/pull/155#discussion_r2212766856) comment we could get rid of this entirely and simply save a foreign key to the oauth instance instead.

users/views.py
```diff
@@ -683,13 +684,14 @@ def post(self, request, *args, **kwargs):
...
-                return JsonResponse({"error_code": ErrorCodes.INVALID_DATA}, status=400)
+            except (IntegrityError, AttributeError):
+                failed_creds.append(index)
+                continue
             phone_numbers = cred.get("users", [])
             users = ConnectUser.objects.filter(phone_number__in=phone_numbers, is_active=True)
             for user in users:
                 UserCredential.add_credential(user, credential, request)
-        return HttpResponse()
+        return JsonResponse({"failed": failed_creds})
```
calellowitz, line 689: I think its nice to send the successful ones as well so the sender can verify (and not just assume), if it wants to.

users/views.py
```diff
@@ -647,13 +648,13 @@ def get(self, request, *args, **kwargs):
...
 
-def get_request_server(request):
+def get_issuing_auth(request):
     auth_header = request.headers.get("authorization")
     encoded_credentials = auth_header.split(" ")[1]
     decoded_credentials = base64.b64decode(encoded_credentials).decode("utf-8")
     client_id, client_secret = decoded_credentials.split(":")
-    server = get_object_or_404(Application, client_id=client_id)
-    return server
+    issuing_auth = get_object_or_404(IssuingAuthority, oauth_application__client_id=client_id)
```
calellowitz, line 656: nit: I don't think you want `get_object_or_404` here since if it doesn't exist, that is really an auth error (their credentials are not fully set up).

calellowitz (approved):
not blocking but a couple suggestions

users/views.py
zandre-eng: [ab48e55](https://github.com/dimagi/connect-id/pull/155/commits/ab48e55b7b7b883db6f07a2c1641de3fd720c10e)

calellowitz (approved):
Approved pending feedback on the model PR

PR #157, opened 2025-07-08
List Credentials Endpoint

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1373?atlOrigin=eyJpIjoiZmU4NTA5ZDM3MmIzNGQ1ZmIyYjQ5ZDAzMzZkOGFhMDUiLCJwIjoiaiJ9).
Link to tech spec [here](https://docs.google.com/document/d/1lHHZEx7_MOWxu1RxH1wXb7wGuLWHHp68wWx-jtWwBX8/edit?tab=t.0#heading=h.hwyiocd7jsqu).

This PR adds a new `list_credentials` GET endpoint as part of the wider changes to support PersonalID Worker History. Essentially, this new endpoint allows for showing all of the collected credentials for a specific user. The endpoint uses basic auth for authentication, and uses the authenticated user object to determine which credential to return.

There are no expected query params for this endpoint. The response payload will be in the following format:
```
{
    "credentials": [
        {
            "uuid": "a95feaab777b41c1b05f7118dc697d9c",  // Randomly generated UUID
            "app_id": "beaabebd004c4b948bd6182230789ef7", // App ID of HQ app if issuer is HQ else None
            "opp_id": "90f85977b80a45589fc1a737833b322f", // Opportunity ID if issuer is CONNECT else None
            "date": "2023-10-01T12:00:00Z", // ISO 8601 format
            "title": "Credential Title",  // Name of the credential / App name
            "issuer": "HQ", // Or CONNECT,
            "issuer_environment": "production", // Or staging/india
            "level": "3_MONTH_ACTIVE",  // Code/Qualifier of the credential
            "type": "LEARN",  // OR DELIVER for Connect OR APP_ACTIVITY for HQ
            "slug": "beaabebd004c4b948bd6182230789ef7"
        },
        ...
    ]
}
```

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -29,7 +29,7 @@ from utils.app_integrity.exceptions import DuplicateSampleRequestError
 from utils.app_integrity.google_play_integrity import AppIntegrityService
 from utils.rest_framework import ClientProtectedResourceAuth
 
-from .auth import ServerKeysAuthentication, SessionTokenAuthentication
+from .auth import IssuingCredentialsAuth, SessionTokenAuthentication
 from .const import NO_RECOVERY_PHONE_ERROR, TEST_NUMBER_PREFIX, ErrorCodes, SMSMethods
 from .exceptions import RecoveryPinNotSetError
 from .fcm_utils import create_update_device
@@ -652,7 +652,7 @@ def get_issuing_auth(request):
 
 
 class AddCredential(APIView):
-    authentication_classes = [ServerKeysAuthentication]
+    authentication_classes = [IssuingCredentialsAuth]
 
     def post(self, request, *args, **kwargs):
         creds = request.data.get("credentials")
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,4 @@
 import base64
-import uuid
 from datetime import timedelta
 
 import pytest
@@ -81,10 +80,10 @@ def credential_issuing_authority():
 
 @pytest.fixture
 def credential_issuing_client(api_client, credential_issuing_authority):
-    secret_key = uuid.uuid4().hex
-    credential_issuing_authority.server_credentials.set_secret_key(secret_key)
-    credential_issuing_authority.server_credentials.save()
-    auth = f"{credential_issuing_authority.server_credentials.client_id}:{secret_key}".encode()
+    auth = (
+        f"{credential_issuing_authority.server_credentials.client_id}:"
+        f"{credential_issuing_authority.server_credentials.secret_key}"
+    ).encode()
     credentials = base64.b64encode(auth).decode("utf-8")
     api_client.defaults["HTTP_AUTHORIZATION"] = "Basic " + credentials
     return api_client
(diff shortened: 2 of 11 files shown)


users/views.py
```diff
@@ -661,6 +662,27 @@ def post(self, request, *args, **kwargs):
         return HttpResponse()
 
 
+class ListCredentials(APIView):
+    authentication_classes = [BasicAuthentication]
+
+    def get(self, request, *args, **kwargs):
+        user_creds = UserCredential.objects.filter(user=request.user).select_related("credential")
```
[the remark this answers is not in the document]
zandre-eng: Great suggestion, we can definitely do that. Updated in [00245be](https://github.com/dimagi/connect-id/pull/157/commits/00245be6cb9d3d33f39501cf847b29bf72212d78).

users/views.py
```diff
@@ -661,6 +662,27 @@ def post(self, request, *args, **kwargs):
         return HttpResponse()
 
 
+class ListCredentials(APIView):
+    authentication_classes = [BasicAuthentication]
+
+    def get(self, request, *args, **kwargs):
+        credentials = Credential.objects.filter(usercredential__user=request.user)
+        results = [
+            {
```
calellowitz, line 671: does this not need to pass the `slug` based on the spec?

users/views.py
```diff
@@ -661,6 +662,27 @@ def post(self, request, *args, **kwargs):
         return HttpResponse()
 
 
+class ListCredentials(APIView):
+    authentication_classes = [BasicAuthentication]
```
calellowitz, line 666: I think the default authentication classes should be fine for this (unless you are intentionally preventing the use of tokens)

zandre-eng: Happy to go with the default authentication classes. I was simply following the spec which outlined using basic authentication here.

sravfeyn (approved):
LGTM, except for the pending comment from Cal

users/views.py
Charl1996: Will mobile use the `slug` for anything?

zandre-eng: After reaching out to mobile it sounds like we won't be using the slug (at least, not for the time being anyways) and will be using the `app_id` and `opportunity_id` fields.

calellowitz (approved):
pending possible changes to the model this looks fine.

calellowitz (approved):

users/views.py
```diff
@@ -684,6 +684,27 @@ def post(self, request, *args, **kwargs):
         return JsonResponse({"success": success_creds, "failed": failed_creds})
 
 
+class ListCredentials(APIView):
+    def get(self, request, *args, **kwargs):
+        credentials = Credential.objects.filter(usercredential__user=request.user)
+        results = [
+            {
+                "uuid": c.uuid,
```
pxwxnvermx, line 692: nit: Can we use DRF Serializers for generating this data from the objects?

pxwxnvermx (approved):
Just a nit. Rest looks good. Thanks

users/views.py
zandre-eng: [d0485bb](https://github.com/dimagi/connect-id/pull/157/commits/d0485bb3695434be833562e2644d1875083c5a5c)

pxwxnvermx (approved):
LGTM. Thanks

PR #159, opened 2025-07-17
Store integrity samples

Charl1996:
## Technical Summary
[Ticket](https://dimagi.atlassian.net/browse/CI-164)

Related [HQ changes](https://github.com/dimagi/commcare-hq/pull/36737).

This PR adds the temporary view needed to report an app's integrity for sampling purposes. 

Currently the view returns whether the check "passed" or "failed", but if more info is needed I'm happy to make updates.

The view also supports test applications and demo users in case we want to test this (might be useful for mobile team). 

## Logging and monitoring
Don't think logging is necessary here.

### Safety story
Added a bunch of tests for this view.

This PR does make small changes to the "AppIntegrityService" class, but the existing tests covers that so I'm confident it won't be a problem in production.

### Automated test coverage
Added new tests and updated existing tests where necessary.

### QA Plan
No QA planned

### Labels & Review


users/views.py
```diff
@@ -865,3 +874,62 @@ def confirm_session_otp(request):
...
+        token=integrity_token,
+        request_hash=request_hash,
+        app_package=app_package,
+        is_demo_user=is_demo_user,
+    )
+
+    raw_verdict = service.obtain_verdict()
+
+    try:
+        sample = DeviceIntegritySample.objects.get(
```
calellowitz, line 909: can this use `get_or_create`?

users/models.py
```diff
@@ -253,6 +253,27 @@ class Meta:
...
 
+class DeviceIntegritySample(models.Model):
+    class ProcessedVerdict(models.TextChoices):
+        PASSED = "passed", "passed"
+        FAILED = "failed", "failed"
+
+    device_id = models.CharField(max_length=255)
+    google_verdict = models.JSONField()
+    processed_response = models.CharField(max_length=100, blank=True)
+    processed_verdict = models.CharField(
```
calellowitz, line 264: nit: this can be a boolean, which will make it easier to use and query against

users/models.py
```diff
@@ -253,6 +253,27 @@ class Meta:
...
 
 
+class DeviceIntegritySample(models.Model):
+    class ProcessedVerdict(models.TextChoices):
+        PASSED = "passed", "passed"
+        FAILED = "failed", "failed"
+
+    device_id = models.CharField(max_length=255)
+    google_verdict = models.JSONField()
+    processed_response = models.CharField(max_length=100, blank=True)
```
calellowitz, line 263: I think rather than storing one column here, we want to store each test result separately as a boolean, so each test in `analyze` verdict is stored separately. That will make it much easier to query and analyze, as well as allow us to tell which fail multiple checks such as unlicensed and app integrity (or device integrity)

users/models.py
```diff
@@ -253,6 +253,27 @@ class Meta:
         constraints = [models.UniqueConstraint(fields=["phone_number", "session"], name="phone_number_session")]
 
 
+class DeviceIntegritySample(models.Model):
+    class ProcessedVerdict(models.TextChoices):
+        PASSED = "passed", "passed"
+        FAILED = "failed", "failed"
+
```
calellowitz, line 260: this will need some kind of guid to identify the user, which can't be device id because those aren't traceable to the user in hq. I thought HQ was going to send one down (the user id in hq), that mobile would send back to us. Device id is good to also have, but we will want to know the hq user as well

calellowitz (changes requested):
Few things (some smaller than others), but I think we are on the right track

users/models.py
Charl1996: I think I'm going to use a JSON field there then, where the key is the type of check performed and the value the boolean outcome

users/models.py
Charl1996: That's fair

Charl1996: https://github.com/dimagi/connect-id/pull/159/commits/0dafc5cef07772600b37c7c6c6124456b0734f91

Charl1996:
I updated a couple of things:
1. The fact that the view did too much and had to know too much of how the service operates bothered me, so I moved the sampling logic into the service.
2. If a particular sampling request has been performed already we don't perform it again.

utils/app_integrity/google_play_integrity.py
```diff
@@ -93,10 +108,56 @@ def _check_device_integrity(self, device_integrity: DeviceIntegrity):
...
+                passed_device_integrity_check = False
+            except AccountDetailsError:
+                passed_account_details_check = False
+
+        check_passed = (
+            passed_request_check
+            and passed_app_integrity_check
+            and passed_device_integrity_check
+            and passed_account_details_check
+        )
```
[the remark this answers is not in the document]
Charl1996: I thought about that, but having a dedicated field makes it very easy to query "failed" or "passed" records.

users/views.py
```diff
@@ -880,12 +880,15 @@ def confirm_session_otp(request):
...
-    external_id = data.get("request_id")
+    request_id = data.get("request_id")
     device_id = data.get("cc_device_id")
 
-    if not (external_id and device_id):
+    if not (request_id and device_id):
         return JsonResponse({"error_code": ErrorCodes.MISSING_DATA}, status=400)
 
+    if DeviceIntegritySample.objects.filter(request_id=request_id).exists():
+        return JsonResponse({"result_code": None}, status=304)
```
calellowitz: what is 304? and does mobile know how to handle this response?

users/views.py
```diff
@@ -909,39 +912,42 @@ def report_integrity(request):
...
 
+    passed_request_check = True
+    passed_app_integrity_check = True
+    passed_device_integrity_check = True
+    passed_account_details_check = True
+
     for evaluator in service.evaluators:
         try:
             evaluator(verdict)
         except IntegrityRequestError:
```
calellowitz: will this result in incorrect information if multiple checks failed? Because it will catch the first exception and then continue on with the other fields all marked as true? They won't even be tested

Charl1996: > will this result in incorrect information if multiple checks failed?

No, it won't

> it will catch the first exception and then continue on with the other fields all marked as true?

Not quite. We're iterating over all checks and evaluating each one, regardless of whether any checks failed before that. In this way we give each check the chance to say something. 

I've added a test now for multiple check failures here, proving the above: https://github.com/dimagi/connect-id/pull/159/commits/5b164b769fdc162eee1594f4dd01d16c54ba3057

users/views.py
Charl1996: 304 indicates that the resource was not modified. After reading up a bit wider I realise this is not the best status code for this particular use-case. I think returning a 200 here with `result_code: None` would be fine.

The intention was that we simply don't do anything if the sample was already taken. Mobile won't theoretically make hit this endpoint twice for the same user, but I thought it's probably prudent to not do anything if there's a sample for this user already, in case mobile does accidentally make the request twice.

Charl1996: https://github.com/dimagi/connect-id/pull/159/commits/1905bfef8749fff532b6e0273187ff93024ab18d

users/views.py
calellowitz: Yep. I misread the indentation. You are correct

PR #166, opened 2025-08-05
Support Fetching Users with Usernames

zandre-eng:
## Technical Summary

Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1372).
Link to tech spec [here](https://docs.google.com/document/d/1lHHZEx7_MOWxu1RxH1wXb7wGuLWHHp68wWx-jtWwBX8/edit?tab=t.0#heading=h.rm3z4rlzvedt).

There is [a PR](https://github.com/dimagi/commcare-hq/pull/36803) for HQ to submit credentials to PersonalID. Currently, HQ does not store the ConnectID user's phone number, and only has their username as the means to reference a ConnectID user. Given this, the endpoint for adding credentials needs to support either usernames or phone numbers for fetching users to assign credentials to.

## Logging and monitoring

No logging requirements.

### Safety story

- Local testing.
- Unit tests exist.

### QA Plan

No QA planned.

### Labels & Review

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -678,11 +678,7 @@ class AddCredential(APIView):
                 failed_creds.append(index)
                 continue
             success_creds.append(index)
-            if "usernames" in cred:
-                users = ConnectUser.objects.filter(username__in=cred["usernames"], is_active=True)
-            else:
-                phone_numbers = cred.get("users", [])
-                users = ConnectUser.objects.filter(phone_number__in=phone_numbers, is_active=True)
+            users = ConnectUser.objects.filter(username__in=cred.get("usernames", []), is_active=True)
             for user in users:
                 UserCredential.add_credential(user, credential, request)
         return JsonResponse({"success": success_creds, "failed": failed_creds})
--- a/users/tests/test_views.py
+++ b/users/tests/test_views.py
@@ -394,7 +394,7 @@ class TestAddCredential:
         payload = {
             "credentials": [
                 {
-                    "users": [user.phone_number.raw_input, "1234567890"],
+                    "usernames": [user.username, "does-not-exist"],
                     "title": "Test Credential",
                     "app_id": app_id,
                     "type": "DELIVER",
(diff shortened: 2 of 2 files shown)


users/views.py
```diff
@@ -678,8 +678,11 @@ def post(self, request, *args, **kwargs):
                 failed_creds.append(index)
                 continue
             success_creds.append(index)
-            phone_numbers = cred.get("users", [])
-            users = ConnectUser.objects.filter(phone_number__in=phone_numbers, is_active=True)
+            if "usernames" in cred:
```
Charl1996, line 681: The `usernames` are being populated in the payload [here](https://github.com/dimagi/commcare-hq/blob/5c319051114aa6639258c170442b5b94ce6e097d/corehq/apps/users/credentials_issuing.py#L124), right?

users/views.py
```diff
@@ -678,8 +678,11 @@ def post(self, request, *args, **kwargs):
...
+                users = ConnectUser.objects.filter(username__in=cred["usernames"], is_active=True)
+            else:
+                phone_numbers = cred.get("users", [])
```
Charl1996, line 684: I don't see `users` ever being populated in the `credentials` payload ([this PR](https://github.com/dimagi/commcare-hq/pull/36803/files#diff-d808e1820eec36aac42dce0d8b4221999143269b813813181b481bc4e27a3bec))?

calellowitz, line 684: should this be `phone_numbers` or similar to be a bit clearer? Otherwise, its not obvious what the difference between a `users` or `usernames` payload would be.

calellowitz (commented):
Do we need to support both username and phone number, or is username alone sufficient?

users/views.py
Charl1996: Great! Thanks for clarifying.

Charl1996:
Looks good to me. 

I do like [Cal's suggestion](https://github.com/dimagi/connect-id/pull/166#discussion_r2258499796) and would like to see that as well, but I suspect that will also need Connect changes.

I'm happy to approve if there's at least a comment in the code for now making it more clear what the difference is between the keys (or how they mean the same, but come from different places).

Charl1996:
@zandre-eng I think that's fine.

users/views.py
zandre-eng: As part of the discussion in this PR, I have changed this so that we only support usernames now in [f4441cd](https://github.com/dimagi/connect-id/pull/166/commits/f4441cd8a3a1e33b8318e4556cb003420f1b9f58).

calellowitz:
> @calellowitz We could support usernames only, although this would require some changes on Connect. Specifically, we'll need to change the form on Connect to support entering usernames instead of phone numbers and then change the field in the payload to send through usernames instead.

Ah, I hadn't realized this was in use in connect, agree the new plan is fine for now, but when we do make the change in Connect I wonder if we even need to change the form. Connect should be able to look up usernames for a number fine (and thats probably easier for the user).

calellowitz (approved):

PR #167, opened 2025-08-08
Add device_id to ConfigurationSession

Charl1996:
## Technical Summary
[Spec](https://docs.google.com/document/d/1wTrmOml_BT27yvq0SznYNbnGLHiskNXxrGFbe8zPTEQ/edit?tab=t.6rvbwosifd3r)
No ticket exists, but since it's such a low hanging fruit, I plucked it.

This PR adds the (optional) device_id to the `ConfigurationSession` model and reads a `device_id` from the `start_configuration` request data when creating a configuration session.

## Logging and monitoring
This is meant for monitoring in a sense.

### Safety story
Checked it locally (using factory and django admin)

### Automated test coverage
No additional test

### QA Plan
No QA planned

### Labels & Review

## Tasks post-merge
Update [API docs](https://docs.google.com/spreadsheets/d/1J0tiUK0Hf7LSCxne-4wwQHOykIsl_Hhaq95B6LtwRec/edit?gid=1351908939#gid=1351908939).


calellowitz (approved):

Charl1996:
@sravfeyn 

> Code looks good, but somehow adding a field to a model just for monitoring doesn't feel great. Are there any other ways you could get that info without editing the model.

I don't see an alternative, less intrusive way of keeping track of the device ID. The only real alternative I see is adding a new table, but that's a worse option for me because we're already tracking all the data we want linked together in the session table and adding a new table seems like we're duplicating data storage.

Any thoughts?

zandre-eng (approved):

PR #175, opened 2025-08-21
Specify OTP strategy in configuration session

Charl1996:
[Ticket](https://dimagi.atlassian.net/jira/software/c/projects/CCCT/boards/276?assignee=606da52fedc14f00769a8a86&selectedIssue=CCCT-1578)
[Tech Spec](https://docs.google.com/document/d/19kNyXbpqy5Y-qQEhKHAgY5Eq1Zod8L7lyNTiTV8lLSA/edit?tab=t.0) (still in review)

> [!NOTE]
> This PR was done because I considered it low hanging fruit, but it is subject to change if the spec changes.

## Technical Summary

This PR adds a new column to the `ConfigurationSession` table, `otp_sms_strategy`, which specifies whether the sessions should send the OTP SMS using firebase or PersonalID. 

## Logging and monitoring
N.A

### Safety story
Tested locally and verified that all configuration sessions have `firebase` specified by default.

### Automated test coverage
No new tests added.

### QA Plan
N.A

### Labels & Review

### DIFF
--- a/users/const.py
+++ b/users/const.py
@@ -32,8 +32,3 @@ class ErrorCodes:
 
 
 MAX_BACKUP_CODE_ATTEMPTS = 3
-
-
-class SMSMethods:
-    PERSONAL_ID = "personal_id"
-    FIREBASE = "firebase"
--- a/users/views.py
+++ b/users/views.py
@@ -30,7 +30,7 @@ from utils.app_integrity.google_play_integrity import AppIntegrityService
 from utils.rest_framework import ClientProtectedResourceAuth
 
 from .auth import IssuingCredentialsAuth, SessionTokenAuthentication
-from .const import NO_RECOVERY_PHONE_ERROR, TEST_NUMBER_PREFIX, ErrorCodes, SMSMethods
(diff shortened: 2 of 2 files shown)


users/models.py
```diff
@@ -265,6 +269,7 @@ class ConfigurationSession(models.Model):
     gps_location = models.CharField(max_length=100, blank=True, null=True)  # GPS coordinates in format "lat lon"
     invited_user = models.BooleanField(default=False)
     device_id = models.CharField(max_length=255, blank=True, null=True)
+    otp_sms_strategy = models.CharField(choices=OTPSMSStrategy.choices, default=OTPSMSStrategy.FIREBASE, max_length=15)
```
sravfeyn, line 272: Why would we want to store this at the session level? Not as some of user level setting that can be updated? (if what I understood from the spec is correct, 'a way to toggle to sms strategy to Twilio in case needed')

Charl1996: Oh, I imagined it would be done at django admin level by a developer.

pxwxnvermx (approved):

PR #178, opened 2025-08-28
ensure that 'channel_name' is never `None`

snopoke:
## Technical Summary
Fixes [CONNECT-ID-2W](https://dimagi.sentry.io/issues/6839375641/)

I'm not sure why this only happens sometimes but there is an error from Firebase which happens sometimes when sending the channel creation message:

`ValueError: Message.data must not contain non-string values.`

Looking at Sentry it appears that the only non-string value is the `channel_name` which is `None`.

OCS is calling this endpoint and never sends `channel_name` in the request so I don't know why this doesn't always fail.

Looking at the code, this change seems more logical that the original code. `channel_source` in the data is set to `channel_source` and `channel_name` is set using `channel.visible_name` which is `channel_name or channel_source`.

## Logging and monitoring
Existing Sentry integration.

### Safety story

### Automated test coverage
Unit tests have been updated

### QA Plan
None

### Labels & Review

- [ ] The set of people pinged as reviewers is appropriate for the level of risk of the change

### DIFF
--- a/messaging/tests.py
+++ b/messaging/tests.py
@@ -140,7 +140,7 @@ def rest_channel_data(user=None, consent=False, channel_name=None):
         "user_consent": consent,
         "connectid": str(user.username) if user else None,
         "channel_source": "hq project space",
-        "channel_name": channel_name
+        "channel_name": channel_name,
     }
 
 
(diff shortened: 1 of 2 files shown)


hemant10yadav (commented):
The test failures seem genuine.

PR #162, opened 2025-07-29
Flag for Manual OTP

zandre-eng:
## Technical Summary

Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1520).
Link to tech spec [here](https://docs.google.com/document/d/1hhHsbgD6PT_bpxp6ZnyyGZbQcHCV7MUjJJG1bGUubOE/edit?tab=t.0).

A new `has_manual_otp` field has been added to the `SessionPhoneDevice` model, along with overriding the `generate_challenge` function to block from generating and sending a new OTP SMS if a manual one has been generated (unless it is about to expire). 

This is the first PR in a series of upcoming work to allow for generating and distributing a manually generated OTP.

## Logging and monitoring

No monitoring requirements as this is simply just some changes to the model.

### Safety story

- Local testing.
- Unit tests exist.

Mostly just adds a new model field, and so should not impact any existing/previous versions on mobile.

- [X] I am confident that this change will not break current and/or previous versions of CommCare apps

### Automated test coverage

Automated tests do exist to test the two views that use the `SessionPhoneDevice` model.

### QA Plan

No QA planned.

### Labels & Review

- [X] The set of people pinged as reviewers is appropriate for the level of risk of the change

PR #181, opened 2025-09-05
Don't change has_manual_otp value back to False

Charl1996:
## Technical Summary
[Ticket](https://dimagi.atlassian.net/browse/CCCT-1507)
[Brief spec](https://docs.google.com/document/d/1UoZMrSNwgRt_EK4lNGxmLXdyn6IEo5uebMDhnWiFjsA/edit?tab=t.0#heading=h.mup0fcuh9wit)

We're interested to see how many sessions have been using the manual OTP workflow.

This PR removes the code which sets the `has_manual_otp` back to `False` once it's set to `True` as the field has been repurposed for "logging only" and does not dictate any OTP generation behaviour as [seen in this comment](https://github.com/dimagi/connect-id/pull/162#discussion_r2242164934).

The [generate_challange method](https://github.com/dimagi/connect-id/blob/fbbea45d45e2f2a5e92c03798360adea2849d2e9/users/models.py#L146C9-L146C27) on the superclass already takes the OTP code livespan into account and does not generate a new OTP if the current one is not close to expiry, so even if we don't consider the status of `has_manual_otp` we still won't generate a new OTP if the user is requesting one and the current one is not close to expiry. 

## Logging and monitoring
No monitoring needed.

### Safety story

### Automated test coverage
Automated tests exists, but no new ones added.

### QA Plan
QA not planned.

### Labels & Review

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -307,15 +307,6 @@ class SessionPhoneDevice(BasePhoneDevice):
     class Meta:
         constraints = [models.UniqueConstraint(fields=["phone_number", "session"], name="phone_number_session")]
 
-    def generate_challenge(self):
-        if self.is_otp_close_to_expiry:
-            # Set to false as the token is close to expiry and
-            # we want to auto generate a new one
-            self.has_manual_otp = False
-            self.save()
-        if not self.has_manual_otp:
-            return super().generate_challenge()
-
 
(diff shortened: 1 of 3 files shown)


zandre-eng (approved):
Thanks for the change!

PR #182, opened 2025-09-10
Enforce uniqueness explicitly

Charl1996:
## Technical Summary
This PR is an attempt to fix [this sentry error](https://dimagi.sentry.io/issues/6866466478/?alert_rule_id=15706404&alert_timestamp=1757403115238&alert_type=email&environment=local&notification_uuid=e96a7bde-57c2-4a85-87ba-95d5443073a8&project=4508576093044736&referrer=alert_email).

It's slightly surprising to me that this error occurred, since we explicitly checked whether the a record with the given request_id already existed...but nonetheless, the current implementation handles this situation at the source, which also accounts for potential race conditions. 

## Logging and monitoring
N.A

### Safety story

### Automated test coverage
Extensive unit tests already exists.

### QA Plan
No QA.

### Labels & Review

### DIFF
--- a/utils/app_integrity/google_play_integrity.py
+++ b/utils/app_integrity/google_play_integrity.py
@@ -120,9 +120,6 @@ class AppIntegrityService:
         """
         Performs a sampling request to log the integrity check results.
         """
-        if DeviceIntegritySample.objects.filter(request_id=request_id).exists():
-            raise DuplicateSampleRequestError("Duplicate sample request")
-
         raw_verdict = self.obtain_verdict()
         verdict = self.parse_raw_verdict(raw_verdict)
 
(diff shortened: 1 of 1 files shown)


zandre-eng:
> also accounts for potential race conditions.

@Charl1996 Could you say more about this? My understanding is that if a specific device hits the view that calls `log_sample_request()` then there shouldn't be a race condition since this device won't hit this endpoint more than once at a time. If it does then I would be surprised and we should maybe check to make sure that this endpoint isn't getting spammed by mobile.

Charl1996:
> My understanding is that if a specific device hits the view that calls log_sample_request() then there shouldn't be a race condition since this device won't hit this endpoint more than once at a time.

Correct, although there could be a bug or something on mobile. It doesn't look like there's multiple such cases on sentry. I'll let mobile know it happened, maybe they can spot something.

users/views.py
```diff
@@ -928,7 +929,8 @@ def report_integrity(request):
             request_id=request_id,
             device_id=device_id,
         )
-    except DuplicateSampleRequestError:
+    except DuplicateSampleRequestError as e:
+        sentry_sdk.capture_exception(e)
```
Charl1996, line 933: So we can track if there's something funky going on...

calellowitz (commented):
left a few comments the code structure one is more important than the sentry one

PR #185, opened 2025-09-19
case insensitive usernames

calellowitz:
## Technical Summary

At some point in the past usernames in connect were switched from upper to lowercase. Since many of the systems we work with are case insensitive and don't preserve the casing of the username this was causing problems setting up channels. This PR updates our channel code to match usernames without case sensitivity. To make that safe, it also adds a constraint preventing two usernames that differ only in case, by enforcing that the lower case version of each username is unique.

### Safety story

### Automated test coverage

I added tests to both uppercase and lowercase user names in the create channel call to confirm both work.

### Labels & Review


pxwxnvermx (approved):
LGTM. Thanks for the fix. 😄

PR #190, opened 2025-10-08
Fix Error Saving Server Keys

zandre-eng:
## Technical Summary

Link to Sentry error [here](https://dimagi.sentry.io/issues/6930711840/?project=4508576093044736&query=is%3Aunresolved&referrer=issue-stream).

This PR fixes a 500 error that happens when a user tries to save a ServerKeys instance using the Django admin panel. The wrong accessor was being used to set the secret key and so a 500 would occur.

Additionally, a small grammar update is made for the `ServerKeys` model so that it displays better in the admin panel.

## Logging and monitoring

Will be able to monitor Sentry to confirm that the error has been resolved.

### Safety story

- Local testing.

### Automated test coverage

N/A

### QA Plan

No QA planned.

### Labels & Review


users/models.py
```diff
@@ -194,6 +194,10 @@ class ServerKeys(models.Model):
     def __str__(self):
         return self.name
 
+    class Meta:
+        verbose_name = "Server Keys"
```
Charl1996, line 198: Curious what the previous display name was?

Charl1996 (approved):

users/models.py
zandre-eng: Django defaults to adding an "s" to make things plural, so in the admin panel it was being displayed as "Server Keyss" (always irked me a little).

calellowitz: Obviously a tiny thing, but that is why the convention is to always name models in the singular

PR #221, opened 2026-03-19
consider demo users as invited

sravfeyn:
## Technical Summary

https://dimagi.atlassian.net/browse/CI-554

## Logging and monitoring

NA

### Safety story

Tests exist and impacts only demo users.

### Automated test coverage

Tests exist
### Labels & Review

### DIFF
--- a/utils/app_integrity/decorators.py
+++ b/utils/app_integrity/decorators.py
@@ -61,7 +61,8 @@ def require_app_integrity(view):
         request_hash = request.headers.get(INTEGRITY_REQUEST_HASH_KEY)
         phone_number = request.data.get("phone_number", "")
 
-        invited = check_number_for_existing_invites(phone_number)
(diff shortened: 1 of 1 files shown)


mkangia (approved):
Can you highlight in description the test that covers this scenario?

pxwxnvermx (approved):

calellowitz (changes requested):
I don't think this is correct behavior. Demo users should work the same as regular users, returning invited if they have been and not otherwise. There is nothing unique about these users that should make it harder to determine that, and this will let us use demo users to test all workflows

shubham1g5:
@sravfeyn what are next steps here ? QA is getting blocked on this repeatedly at the moment and would like for this to be prioritized soon

mkangia:
Thanks for flagging this @shubham1g5 

@sravfeyn is offline so I have asked @hemant10yadav to take a look.

PR #227, opened 2026-05-15
fix: celery beat fails to create schedule file in /app

ajeety4:
## Technical Summary

`celery beat` is failing on production with `_gdbm.error: [Errno 13] Permission denied: 'celerybeat-schedule'`, so beat exits at startup and **every periodic task in `app.conf.beat_schedule` silently stops firing**. The worker stays up; ad-hoc tasks still work.

**Root cause.** In our Dockerfile, `WORKDIR /app` runs while still root, so `/app` itself is created as `root:root 755`. `COPY --chown=django:django . /app/` only chowns the *contents*, not `/app` itself. After `USER django`, the runtime user can read `/app` but cannot create files in it. Celery's default `PersistentScheduler` writes `celerybeat-schedule` to CWD (`/app`) → EACCES.

**Fix.** Migrate to `django-celery-beat` and use its `DatabaseScheduler`, which stores schedule state in Postgres and **does not touch the filesystem at all**. This matches the pattern used in sibling repo `commcare-connect` (`config/settings/base.py` → `CELERY_BEAT_SCHEDULER = "django_celery_beat.schedulers:DatabaseScheduler"`) and resolves the divergence the original PR description called out.

Changes in this PR:
- Add `django-celery-beat` to `requirements.in` and register `django_celery_beat` in `INSTALLED_APPS`.
- Set `CELERY_BEAT_SCHEDULER = "django_celery_beat.schedulers:DatabaseScheduler"` in `connectid/settings.py`.
- Drop the in-memory `app.conf.beat_schedule` dict from `connectid/celery_app.py` (DatabaseScheduler ignores it).
- Seed the 3 periodic tasks via data migration `messaging/migrations/0008_periodic_tasks.py` so deploys come up with the canonical schedule. Uses `update_or_create` keyed on task name — safe to re-run on existing rows.
- No `--schedule` flag or `/var/run/celery` directory needed; `docker/start_celery` is unchanged from main.

**Restart behavior.** With DatabaseScheduler, schedule state (`last_run_at`, `total_run_count`) is in Postgres, so restarts no longer delay interval tasks — beat resumes from the persisted state. Crontab tasks (`upload_connect_users_to_superset` at midnight) are unaffected by either approach.

## Logging and monitoring

After deploy, in CloudWatch (`connect-id` log group, celery stream):
- `beat: Starting...` with no `Permission denied` traceback.
- `Scheduler: Sending due task resend_notifications_for_undelivered_messages (...)` within 1h (fastest signal).
- Daily `delete_old_messages` and next-midnight `upload_connect_users_to_superset` lines.

Schedule state can be inspected directly in Postgres:

```sql
SELECT name, task, last_run_at, total_run_count FROM django_celery_beat_periodictask;
```

### Safety story

End-to-end verified locally against a freshly-built image, with the dockerized broker (`redis://redis:6379/0`) and db:

1. Built image, ran migrations — migration `0008_periodic_tasks` seeded the 3 expected `PeriodicTask` rows (verified via `psql`).
2. Added a temporary 1-minute `heartbeat_test` PeriodicTask + matching task; reverted before commit.
3. Ran `celery -A connectid.celery_app worker -B -l INFO` in the rebuilt container against the redis broker.
4. Logs show `beat: Starting...`, `Connected to redis://redis:6379/0`, `Scheduler: Sending due task heartbeat_test (messaging.tasks.heartbeat_test)`, and the task firing exactly on its 1-minute interval (two consecutive fires observed at +60s).
5. After cleanup, the 3 production PeriodicTask rows are present, no `heartbeat_test` row remains.

Implicit confirmation that `DatabaseScheduler` is in use: `heartbeat_test` was only registered in the DB (no `CELERY_BEAT_SCHEDULE` entry); the default in-memory `PersistentScheduler` would have found nothing.

### Automated test coverage

None added. Beat itself is not exercised by the pytest suite (`CELERY_TASK_ALWAYS_EAGER = True` short-circuits dispatch). The migration is a data migration; correctness was verified by running it forward and reverse against a real postgres and observing the resulting row state.

### QA Plan

- After deploy: confirm `beat: Starting...` with no traceback.
- Within ~1h: confirm `Sending due task resend_notifications_for_undelivered_messages`.
- Within 24h: confirm `delete_old_messages` runs.
- Next 00:00 UTC: confirm `upload_connect_users_to_superset` runs.
- Optional spot-check: `SELECT name, last_run_at, total_run_count FROM django_celery_beat_periodictask;` should show non-null `last_run_at` and increasing counts on the interval tasks.

### Labels & Review


sravfeyn:
Thanks for catching this, very much an 'edge' case.

> Migrating to django-celery-beat can be considered if reasonable, however this PR unblocks the issue

Any reason not to do this right away? This will have more parity with commcare-connect which will be nice.

ajeety4:
> Thanks for catching this, very much an 'edge' case.
> 
> > Migrating to django-celery-beat can be considered if reasonable, however this PR unblocks the issue
> 
> Any reason not to do this right away? This will have more parity with commcare-connect which will be nice.

Mainly scope - wanted to keep this PR focused on unblocking prod. Agreed that parity with commcare-connect is the right end state. I am happy to do it right away too considering this issue is not a major blockage.

ajeety4:
Approach updated: switched from the `/var/run/celery` workaround to `django-celery-beat` with `DatabaseScheduler`. Branch rebased — see updated description.

sravfeyn (approved):
Thank you!

hemant10yadav (approved):
Thank you for adding `django-celery-beat`!

requirements/requirements.txt
```diff
@@ -8,8 +8,6 @@ amqp==5.3.1
     # via kombu
 asgiref==3.6.0
     # via django
-async-timeout==5.0.1
```
ajeety4, line 11: Noting that this is no longer needed for python version 3.11 (our docker version), and hence was omitted by pip-compile

PR #229, opened 2026-05-18
feat: configure celery beat with file-based schedule

ajeety4:
## Technical Summary

Replaces the django-celery-beat / `DatabaseScheduler` approach from #227 (reverted via #228) with a simpler file-based schedule.

- Defines `CELERY_BEAT_SCHEDULE` in `settings.py` for the three periodic tasks (same intervals as the seed migration in #227): `delete_old_messages` (24h), `resend_notifications_for_undelivered_messages` (1h), `upload_connect_users_to_superset` (daily at midnight UTC).
- Updates `docker/start_celery` to pass `--schedule=/var/run/celery/celerybeat-schedule` so beat persists its state to a writable location.
- Creates `/var/run/celery` (owned by the `django` user) in the Dockerfile so the schedule file can be written. The earlier file-based attempt failed because `/app` was not writable; `/var/run/celery` avoids that.

No new dependencies, no database migration, no requirements changes.

## Logging and monitoring

- Celery beat already logs schedule pickup and task dispatch via stdout → docker logs → CloudWatch (existing pipeline).
- The three periodic task functions themselves are unchanged, so their existing logging continues to work.
- No new alerts required.

### Safety story

- Built the image locally (`connectid:local`) and ran `/start_celery` and `/start_migrate` against local Postgres 15 + Redis in parallel.
  - Web container: migrations applied, gunicorn booted, `/users/` returned 200 "pong" in ~130ms.
  - Celery container: beat started, worker connected to Redis, all three tasks discovered, `/var/run/celery/celerybeat-schedule` created and owned by the `django` user.
- Configuration-only change. Doesn't read from or write to existing app tables. Existing tasks and views are untouched.
- Rollback is a simple `kamal rollback <previous-sha>` — there is no DB state to clean up.

### Automated test coverage

No new unit tests — this is config plus a Dockerfile / start script change with no application logic. Existing tests cover the three periodic task functions themselves.

### QA Plan

- Deploy to QA / staging.
- Confirm the celery container starts cleanly and `/var/run/celery/celerybeat-schedule` is created (`docker exec ... ls -la /var/run/celery`).
- Confirm the web container's healthcheck stays green during and after deploy.
- Optionally observe `resend_notifications_for_undelivered_messages` firing at the next hour boundary.

### Labels & Review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,8 @@ RUN chown django /entrypoint /start*
 RUN python /app/manage.py collectstatic --noinput
 RUN chown django:django -R staticfiles
 
+RUN mkdir -p /var/run/celery && chown django:django /var/run/celery
+
 USER django
 
 EXPOSE 8000
--- a/connectid/settings.py
+++ b/connectid/settings.py
@@ -9,6 +9,7 @@ https://docs.djangoproject.com/en/4.1/topics/settings/
 For the full list of settings and their values, see
 https://docs.djangoproject.com/en/4.1/ref/settings/
 """
+
 import logging
 from pathlib import Path
 
--- a/deploy/config/deploy.yml
+++ b/deploy/config/deploy.yml
@@ -21,6 +21,8 @@ servers:
     cmd: /start_celery
     labels:
       traefik.enable: false
+    volumes:
+      - 'connectid-celerybeat:/var/run/celery'
 
 registry:
   server: 037129986032.dkr.ecr.us-east-1.amazonaws.com
--- a/deploy/config/deploy.yml
+++ b/deploy/config/deploy.yml
@@ -18,11 +18,10 @@ servers:
       - 54.160.64.28
     options:
       env-file: '/home/connect/www/docker.env'
+      volume: 'connectid-celerybeat:/var/run/celery'
     cmd: /start_celery
     labels:
       traefik.enable: false
-    volumes:
-      - 'connectid-celerybeat:/var/run/celery'
 
 registry:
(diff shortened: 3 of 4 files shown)


Charl1996 (approved):

snopoke (commented):
FYI: Using a file based schedule without a Docker volume configured will wipe the schedule on every deploy which may lead to unexpected consequences.

ajeety4:
> FYI: Using a file based schedule without a Docker volume configured will wipe the schedule on every deploy which may lead to unexpected consequences.

Yes, per my understanding this only affects the two interval tasks (1h resend_notifications_for_undelivered_messages, 24h delete_old_messages): on a wipe, last_run_at gets reset to startup time, so frequent deploys can delay or skip a fire. 
Agreed that **named volume** is the safer approach, so I added one for `/var/run/celery`.

snopoke (commented):
lgtm

zandre-eng (approved):
Looks good, but curious if you found out why Connect is able to successfully use `DatabaseScheduler` but not PersonalID?

hemant10yadav (approved):

ajeety4:
> Looks good, but curious if you found out why Connect is able to successfully use `DatabaseScheduler` but not PersonalID?

I don't think it s a django-celery-beat issue (works locally) however could be related to the prod env/infra. The deploy failed because gunicorn workers in the new container hung on their first request ; we couldn't isolate why. Moving to file scheduler for now to unblock as it is a lighter change, no new app or migrations.

calellowitz (approved):

PR #231, opened 2026-05-19
Cs/CI-645 - fix message status updates

Charl1996:
## Technical Summary
As part of investigating https://dimagi.atlassian.net/browse/CI-645 I came across a bug in `send_messages_to_service_and_mark_status` where messages' status would never get updated to the status specified.

This PR does two things:
1. Fix a bug in `send_messages_to_service_and_mark_status` where the message status would not get updated (to the status specified as an argument to the function) due to the fact that the response object was compared with the success status code instead of the `response.status_code`. This means the condition never evaluated to True which means the messages were never pushed to the list used to update their status to the specified one in the arguments.
2. Improve the `UpdateReceivedView` by only considering messages to be updated to `CONFIRMED_RECEIVED` which does not already have that status. This also means we're excluding the possibility for these messages to be sent to the message server (via `send_messages_to_service_and_mark_status`) multiple times if the mobile client accidentally hit this endpoint twice.  

## Logging and monitoring
No logging needed for this PR

### Safety story

### Automated test coverage
Automated tests added.

### QA Plan
QA not planned apart from the automated tests added.

### Labels & Review

### DIFF
--- a/messaging/tasks.py
+++ b/messaging/tasks.py
@@ -59,7 +59,7 @@ def send_messages_to_service_and_mark_status(channel_messages, status_to_be_upda
                 },
                 secret=channel.server.server_credentials.secret_key,
             )
-            if response == status.HTTP_200_OK:
+            if response.status_code == status.HTTP_200_OK:
                 sent_message_ids.extend(msg["message_id"] for msg in messages)
 
         except CommCareHQAPIException:
--- a/messaging/tests.py
+++ b/messaging/tests.py
@@ -604,6 +604,22 @@ class TestUpdateReceivedView:
         )
         assert msg_status == MessageStatus.CONFIRMED_RECEIVED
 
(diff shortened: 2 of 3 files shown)


calellowitz:
Can you update the PR description to something more descriptive of the change it contains

Charl1996:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Charl1996:
> Can you update the PR description to something more descriptive of the change it contains

@calellowitz Does the description make more sense now?

pxwxnvermx (approved):

sravfeyn (approved):

PR #232, opened 2026-05-22
Upload ConfigurationSession to Bigquery and Superset

sravfeyn:
## Technical Summary

Ticket https://dimagi.atlassian.net/browse/CCCT-2415

- Adds task to upload ConfigurationSession to Bigquery and Superset
- Refactor for DRYness and uploading one source to multiple destinations

For deploying, we will need to add relevant Google Cloud credentials to env
- Create a service account scoped to the single dataset
- Download the JSON key file
- Set it in the env

## Logging and monitoring

Logging messages are added

### Safety story

Tested on a local account

### Automated test coverage

Added

### QA Plan

Tested locally, will review once deployed.

### Labels & Review


connectid/settings.py
```diff
@@ -352,6 +340,13 @@ class APIVersion:
 GA_API_SECRET = env("GA_API_SECRET", default="")
 FIREBASE_APP_ID = env("FIREBASE_APP_ID", default="")
 
+# Bigquery upload settings
+GOOGLE_APPLICATION_CREDENTIALS = json.loads(os.environ["GOOGLE_CREDENTIALS_JSON"])
+BIGQUERY_DATASET_ID = env("BIGQUERY_DATASET_ID", default="")
+BIGQUERY_CONFIGURATION_SESSION_TABLE = env(
+    "BIGQUERY_CONFIGURATION_SESSION_TABLE", default="personalid_configuration_session"
+)
```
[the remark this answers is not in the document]
sravfeyn: Fixed in c317416. Changed to use `env("GOOGLE_CREDENTIALS_JSON", default="")` with conditional `json.loads` so environments without BigQuery configured can boot normally.

users/tasks.py
```diff
@@ -30,125 +33,165 @@
...
+        try:
+            tmp = tempfile.NamedTemporaryFile(mode="w", newline="", suffix=".csv", delete=False)
+            writer = csv.DictWriter(tmp, fieldnames=self.fields)
+            writer.writeheader()
+            for row in self.queryset.values(*self.fields).iterator(chunk_size=500):
+                writer.writerow({k: self._serialize(row[k]) for k in self.fields})
+        except Exception:
+            tmp.close()
+            Path(tmp.name).unlink(missing_ok=True)
+            raise
```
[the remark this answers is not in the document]
sravfeyn: Fixed in c317416. Added `tmp = None` before the try block and guarded cleanup with `if tmp is not None`.

users/tasks.py
```diff
@@ -30,125 +33,165 @@
...
-        self.table_name = config.get("table_name", "")
         self.schema = config.get("table_schema", "")
+        self.table_name = table_name
 
-    def run(self):
+    def upload(self, csv_path: Path):
         if not self.base_url or not self.username or not self.password:
-            # Upload not configured
             return
+        self._authenticate()
```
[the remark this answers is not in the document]
sravfeyn: Fixed in c317416. Both `upload()` methods now return `bool`, and the tasks log the actual upload status (e.g. `bigquery=True, superset=False`) rather than an unconditional success message.

users/tests/test_tasks.py
```diff
@@ -41,57 +41,77 @@ def test_write_csv_outputs_expected_rows(settings, superset_config):
...
     assert list(row.keys()) == CONNECT_USER_DUMP_FIELDS
 
 
-def test_upload_posts_csv_to_superset(tmp_path, settings, superset_config):
+@pytest.mark.django_db
+def test_csv_generator_raises_when_max_rows_exceeded():
+    UserFactory.create_batch(3)
+    with pytest.raises(Exception, match="exceeds limit"):
+        with CSVGenerator(ConnectUser.objects.all(), CONNECT_USER_DUMP_FIELDS, max_rows=2):
+            pass
```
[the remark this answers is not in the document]
sravfeyn: Fixed in c317416. `CSVGenerator` now raises `ValueError` and the test asserts `pytest.raises(ValueError, ...)`.

connectid/celery_app.py
```diff
@@ -31,4 +31,8 @@
         "task": "users.tasks.upload_connect_users_to_superset",
         "schedule": crontab(hour=0, minute=0),
     },
+    "upload_configuration_sessions": {
+        "task": "users.tasks.upload_configuration_sessions",
+        "schedule": crontab(hour=2, minute=0),
```
calellowitz, line 36: Out of curiosity, why 2AM?

connectid/settings.py
```diff
@@ -352,6 +339,14 @@ class APIVersion:
 GA_API_SECRET = env("GA_API_SECRET", default="")
 FIREBASE_APP_ID = env("FIREBASE_APP_ID", default="")
 
+# Bigquery upload settings
+_google_credentials_json = env("GOOGLE_CREDENTIALS_JSON", default="")
```
calellowitz, line 345: Big json strings in our env file are typically harder to read and safely modify than individual elements. Is there a reason to make this change?

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
+    except Exception as exc:
+        sentry_sdk.capture_exception(exc)
+        logger.exception("Failed to upload ConfigurationSession dump")
+        raise
 
 
 @shared_task(name="users.tasks.upload_connect_users_to_superset")
 def upload_connect_users_to_superset():
-    exporter = ConnectUserSupersetExporter()
+    config = getattr(settings, "SUPERSET_UPLOAD_CONFIG", {})
```
calellowitz: Since `SUPERSET_UPLOAD_CONFIG` is in settings, I don't think we need to use `getattr` here, we can just access it directly, which is a bit clearer to read

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
+    table_name = config.get("table_name", "connect_user_dump")
```
calellowitz: rather than allow this to be empty, can we just make sure this is set in settings

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
+        self._path = Path(tmp.name)
+        return self._path
+
+    def __exit__(self, *args):
+        if self._path:
+            try:
+                self._path.unlink()
+            except FileNotFoundError:
+                pass
+            except Exception:
```
calellowitz: why catch this, rather than just let it fail? Generally catching a bare exception isn't good practice. Is there a more specific one you are trying to prevent?

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
         data = self._request(
             "post",
             "/api/v1/security/login",
             json={"username": self.username, "password": self.password, "provider": "db", "refresh": True},
         )
         token = data.get("access_token")
         if not token:
-            raise SuperusetUserUploadException("Superset login failed: missing access_token")
-        return token
+            raise Exception("Superset login failed: missing access_token")
```
calellowitz: why change from the specific exception to the generic here?

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
-        if self.schema:
-            payload["schema"] = self.schema
-        with csv_path.open("rb") as csv_file:
-            self._request(
-                "post",
-                f"/api/v1/database/{self.database_id}/upload/",
-                data=payload,
-                files={"file": (csv_path.name, csv_file, "text/csv")},
-            )
+            raise Exception("Superset CSRF token missing")
```
calellowitz: same question about specific exceptions

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
+def upload_configuration_sessions():
+    table_name = settings.BIGQUERY_CONFIGURATION_SESSION_TABLE
+    try:
+        with CSVGenerator(
+            ConfigurationSession.objects.all(), CONFIGURATION_SESSION_DUMP_FIELDS, max_rows=500000
+        ) as csv_path:
+            bq_uploaded = BigQueryUploader(table_name).upload(csv_path)
+            superset_uploaded = SupersetUploader(table_name).upload(csv_path)
+        logger.info("ConfigurationSession upload finished (bigquery=%s, superset=%s)", bq_uploaded, superset_uploaded)
+    except Exception as exc:
```
calellowitz: why catch the exception and then reraise it to sentry rather than just let the normal sentry exception reporting do its thing?

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
+        sentry_sdk.capture_exception(exc)
+        logger.exception("Failed to upload ConfigurationSession dump")
+        raise
 
 
 @shared_task(name="users.tasks.upload_connect_users_to_superset")
 def upload_connect_users_to_superset():
```
calellowitz, line 190: I think this needs to go to bigquery as well

connectid/celery_app.py
```diff
@@ -31,4 +31,8 @@
         "task": "users.tasks.upload_connect_users_to_superset",
         "schedule": crontab(hour=0, minute=0),
     },
+    "upload_configuration_sessions": {
```
calellowitz, line 34: can these two tasks be combined into one?

calellowitz (commented):
left a few questions (mostly around exception handling)

users/tasks.py
```diff
@@ -144,6 +147,101 @@ def _serialize_value(value):
...
+
+        csv_path = None
+        try:
+            csv_path = self.generate_csv()
+            self.upload(csv_path)
+            logger.info("Uploaded ConfigurationSession csv to BigQuery")
+        finally:
+            if csv_path:
+                try:
+                    csv_path.unlink()
```
Charl1996: Nit: I like the way you specified `unlink(missing_ok=True)` below, which I presume supress the `FielNotFoundError`. Since we're not doing anything anycase when the `FileNotFoundError` is raised doing a similar thing here seems OK.

Charl1996 (approved):
Looks good generally.

users/tasks.py
```diff
@@ -30,125 +33,169 @@
     "hq_sso_date",
 ]
 
+CONFIGURATION_SESSION_DUMP_FIELDS = [
```
zandre-eng, line 35: Nit: There exists a `const.py` file for the `users` app. How do you feel about moving this there?

users/tasks.py
```diff
@@ -30,125 +33,169 @@
...
         self.username = config.get("username", "")
         self.password = config.get("password", "")
         self.database_id = config.get("database_id")
-        self.table_name = config.get("table_name", "")
         self.schema = config.get("table_schema", "")
+        self.table_name = table_name
 
-    def run(self):
+    def upload(self, csv_path: Path) -> bool:
         if not self.base_url or not self.username or not self.password:
```
zandre-eng: Nit: Not part of the changes so feel free to ignore, but since other parts of this function have been refactored this can also be refactored to `if not (self.base_url and self.username and self.password)`.

zandre-eng (approved):
A few minor nits, but generally looks fine. I think some of the open comments are worth addressing.

connectid/celery_app.py
sravfeyn: Why is that necessary?

users/tasks.py
sravfeyn: Oh, is it. This was not part of ticket's asks. But I could see it being useful, shall I check on the ticket with mobile team or add it right away. With the refactors I did this would be a single line change `BigQueryUploader(table_name).upload(csv_path)`

connectid/settings.py
sravfeyn: This credential file gets directly downloaded from Google cloud account as a JSON file, so it was easier to just directly make use of that JSON, instead of dissecting it and inserting individual key/pairs in the env.

connectid/celery_app.py
sravfeyn: I wanted to keep a bit of a buffer as the CSV upload files are currently stored in the instance (as opposed to AWS S3), so that the file from the other task that runs up before this gets cleaned up.

users/tasks.py
sravfeyn: Done.

users/tasks.py
sravfeyn: Done.

users/tasks.py
sravfeyn: My intention with the bare exception catching and logging was so that the celery logs have a clear message rather than an unhandled exception log. But yeah, I think it doesn't matter much, so I have fixed it.

users/tasks.py
sravfeyn: This was a mistake, reverted it. Thanks for the catch.

users/tasks.py
sravfeyn: Answered [above](https://github.com/dimagi/connect-id/pull/232#discussion_r3312075120)

users/tasks.py
sravfeyn: Answered [above](https://github.com/dimagi/connect-id/pull/232#discussion_r3312075120)

sravfeyn:
@calellowitz Thanks for the review, I have updated the code and responded.

users/tasks.py
calellowitz: no need to ask the mobile team. worst case they don't use it, but i have talked to dave about this in the past and he said we wanted it

connectid/settings.py
calellowitz: ah, makes sense

connectid/celery_app.py
calellowitz: It just seems easier to have one task instead of multiple. Guarantees we do it sequentially and clean up between, etc. Basically your point about scheduling this later to make sure the other completes becomes unnecessary if they are in the same task. It's not a big deal either way, just seems a bit simpler

users/tasks.py
sravfeyn: This is done, btw.

zandre-eng (approved):

calellowitz (approved):
I think the comment about uploading users to bigquery is still pending, but that can be addressed in a followup PR

PR #242, opened 2026-06-02
Email Verification Model Changes

zandre-eng:
## Technical Summary

[CCCT-2433](https://dimagi.atlassian.net/browse/CCCT-2433)
This is a release path 3 feature — Experiments & early stage iterations of product area

This PR adds the data model and email delivery infrastructure for the email OTP verification feature. It introduces a `BaseOTPDevice` abstract class that consolidates the exponential backoff OTP logic currently duplicated across phone device classes, then builds `BaseEmailOTPDevice`, `UserEmailOTPDevice`, and `SessionEmailOTPDevice` on top of it — mirroring the existing `BasePhoneDevice` / `PhoneDevice` / `SessionPhoneDevice` hierarchy. `RateLimitedError` is introduced in `users/exceptions.py` so callers can distinguish a rate-limited OTP resend from other failures. No endpoints or UI changes are included; this PR is the model layer only.

- **`BaseOTPDevice(SideChannelDevice)`** — new abstract base replacing direct `SideChannelDevice` inheritance for all OTP device models; owns `otp_last_sent`, `attempts`, `is_otp_close_to_expiry`, and `_attempt_send()` (exponential backoff, raises `RateLimitedError` on violation)
- **`BasePhoneDevice` refactored** — now inherits from `BaseOTPDevice`; removes duplicated fields and backoff logic; `generate_challenge()` silently swallows `RateLimitedError` to preserve existing behaviour
- **`BaseEmailOTPDevice(BaseOTPDevice)`** — abstract; owns `email` field, `_send_otp()` (calls `send_email_otp_message`), and `generate_challenge()` (propagates `RateLimitedError` to callers)
- **`UserEmailOTPDevice`** — concrete; non-nullable `user` FK; unique constraint on `(user, email)`
- **`SessionEmailOTPDevice`** — concrete; non-nullable `session` FK; overrides `SideChannelDevice`'s inherited `user` FK to nullable (same pattern as `SessionPhoneDevice`); unique constraint on `(session, email)`
- **`ConfigurationSession.verified_email`** — new nullable `EmailField`; holds a verified email during sign-up before a `ConnectUser` is created
- **`ConnectUser` email uniqueness** — new partial unique constraint (`email_active_user`) on active users with non-blank email, consistent with `phone_number_active_user`
- **`send_email_otp_message()`** — thin wrapper around Django's `send_mail`; logs masked domain for observability; subject `"Your PersonalID verification code"`

## Logging and monitoring

- `users/email_utils.py` logs `INFO` on every OTP send with the masked recipient domain (`***@example.com`) for correlation without PII exposure
- `RateLimitedError` surfaces immediately to the caller so the view layer can respond with HTTP 429 — no silent failures

### Safety story

No new endpoints or user-facing behaviour in this PR — it is data model and infrastructure only. The new OTP device tables are empty until the view layer (next PR) begins writing to them. The `email_active_user` unique constraint is partial (active users, non-blank email only), matching the existing `phone_number_active_user` pattern so existing users with blank emails are unaffected.

### Automated test coverage

- **`TestSendEmailOtpMessage.test_sends_correct_subject_body_and_recipient`** — verifies subject, token, validity minutes, and recipient are passed to `send_mail` correctly
- **`TestSendEmailOtpMessage.test_delivery_failure_propagates`** — verifies `SMTPException` from `send_mail` is not swallowed
- **`TestUserEmailOTPDevice.test_email_unique_per_user`** — verifies the `(user, email)` unique constraint raises `IntegrityError` on duplicate
- **`TestSessionEmailOTPDevice.test_email_unique_per_session`** — verifies the `(session, email)` unique constraint raises `IntegrityError` on duplicate
- **`TestConfigurationSessionVerifiedEmail`** — verifies new field defaults to `None` and persists correctly
- **`TestConnectUserEmailUniqueConstraint.test_email_uniqueness_constraint`** — parametrized: two active users same email (error), two active users blank email (ok), two inactive users same email (ok)

### QA Plan

- [ ] Run `pytest users/tests/test_models.py users/tests/test_email_utils.py -v` and confirm all tests pass
- [ ] Confirm `./manage.py migrate` applies migration `0031` cleanly
- [ ] Confirm `ConnectUser` records with blank emails do not violate the new `email_active_user` constraint
- [ ] In Django shell, confirm `UserEmailOTPDevice` and `SessionEmailOTPDevice` can be created with valid owners and reject duplicates

### Labels & Review

[CCCT-2433]: https://dimagi.atlassian.net/browse/CCCT-2433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


users/email_utils.py
```diff
@@ -0,0 +1,14 @@
+import logging
+
+from django.conf import settings
+from django.core.mail import send_mail
+
+logger = logging.getLogger(__name__)
+
+
+def send_email_otp_message(email: str, token: str, validity_minutes: int) -> None:
+    masked = f"***@{email.split('@')[1]}"
```
Charl1996: Any reason we're masking the email user entirely instead of e.g.
> c***@gmail.com

?

users/tests/test_email_utils.py
```diff
@@ -0,0 +1,25 @@
...
+        assert args[0] == "Your PersonalID verification code"
+        assert "123456" in args[1]
+        assert "30" in args[1]
+        assert args[3] == ["user@example.com"]
+
+    @patch("users.email_utils.send_mail")
+    def test_delivery_failure_propagates(self, mock_send_mail):
+        mock_send_mail.side_effect = SMTPException("connection refused")
+        with pytest.raises(SMTPException):
+            send_email_otp_message("user@example.com", "654321", 30)
```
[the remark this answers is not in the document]
zandre-eng: Declining — doesn't apply here. `@skip_app_integrity_check` only matters for tests hitting an API endpoint through the app-integrity decorator; `TestSendEmailOtpMessage` calls `send_email_otp_message()` directly with `send_mail` mocked, so there's nothing to bypass. It's also a _function_ decorator, so on a class it'd replace the class rather than propagate to the methods. Leaving as-is.

users/tests/test_models.py
```diff
@@ -22,3 +24,54 @@ def test_fields(self):
...
+            ("shared@example.com", False, False),  # two inactive users, same email → ok
+        ],
+    )
+    def test_email_uniqueness_constraint(self, email, is_active, should_raise):
+        UserFactory(email=email, is_active=is_active)
+        if should_raise:
+            with pytest.raises(IntegrityError):
+                UserFactory(email=email, is_active=is_active)
+        else:
+            UserFactory(email=email, is_active=is_active)  # must not raise
```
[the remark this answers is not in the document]
zandre-eng: Declining — same reason as the `test_email_utils.py` finding. All four classes only exercise ORM operations directly (`objects.create`, `save`, `refresh_from_db`); none make HTTP requests or pass through a view, so app integrity never runs. All 10 tests already pass without the decorator. `@skip_app_integrity_check` is also a function decorator and can't be applied to a class as described.

users/models.py
```diff
@@ -144,29 +150,50 @@ class Meta:
...
-                sender = get_sms_sender(self.phone_number.country_code)
-                send_sms(self.phone_number.as_e164, message, sender)
+        if self.otp_last_sent is None or now() - self.otp_last_sent >= timedelta(minutes=wait_time):
+            self._send_otp()
             self.otp_last_sent = now()
             self.attempts += 1
             self.save()
+        else:
+            retry_after = int((timedelta(minutes=wait_time) - (now() - self.otp_last_sent)).total_seconds())
+            raise RateLimitedError(retry_after)
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [9cdadf2](https://github.com/dimagi/connect-id/pull/242/commits/9cdadf290e45aa23399abea2614a7762a32ed319).

users/email_utils.py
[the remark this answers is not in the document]
zandre-eng: Addressed in [9cdadf2](https://github.com/dimagi/connect-id/pull/242/commits/9cdadf290e45aa23399abea2614a7762a32ed319).

zandre-eng: No particular reason. I simply defaulted to masking the entire sensitive part, but I can see how there could be value for debugging by having the first character visibile.

Made the first character visible in [9cdadf2](https://github.com/dimagi/connect-id/pull/242/commits/9cdadf290e45aa23399abea2614a7762a32ed319).

users/models.py
```diff
@@ -144,29 +150,59 @@ class Meta:
...
+                self.attempts = 0
+            wait_time = 2**self.attempts
+            if self.otp_last_sent is None or now() - self.otp_last_sent >= timedelta(minutes=wait_time):
+                self._send_otp()
+                self.otp_last_sent = now()
+                self.attempts += 1
+                self.save()
+            else:
+                retry_after = int((timedelta(minutes=wait_time) - (now() - self.otp_last_sent)).total_seconds())
+                raise RateLimitedError(retry_after)
```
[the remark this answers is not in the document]
zandre-eng: Splitting this into its two sub-findings:

**1. `_send_otp()` inside the transaction** — real concern, but the proposed fix isn't safe here. Moving `save()` before `_send_otp()` means a network failure leaves the DB with an already-incremented `attempts` and `otp_last_sent` — the user's backoff clock advances even though they never received a code. The clean solution (`on_commit` hook) doesn't propagate exceptions, which breaks `generate_challenge`'s return-value contract. Noting it as a known tradeoff on this branch rather than applying a half-correct fix.

**2. `is_otp_close_to_expiry` misfiring for short validity** — not valid. Both callers pass `valid_secs=1800` (30 min); the 5-minute threshold never fires spuriously at current settings. Skipping as hypothetical.

users/email_utils.py
```diff
@@ -0,0 +1,15 @@
+import logging
+
+from django.conf import settings
+from django.core.mail import send_mail
+
+logger = logging.getLogger(__name__)
+
+
+def send_email_otp_message(email: str, token: str, validity_minutes: int) -> None:
+    local, domain = email.rsplit("@", 1) if "@" in email else (email, "invalid")
```
pxwxnvermx: We should fail at invalid emails. Trying to send an email to a malformed mail id would fail further down the line so catching it earlier reduces that 1 more call we make to AWS.
An error log with a bare return would also work here.

users/models.py
```diff
@@ -334,6 +371,42 @@ class Meta:
...
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["user", "email"], name="user_email_otp_device_user_email"),
+        ]
+
+
+class SessionEmailOTPDevice(BaseEmailOTPDevice):
+    session = models.ForeignKey(ConfigurationSession, on_delete=models.CASCADE)
+    # non-nullable on the base SideChannelDevice — override to nullable for session-owned devices
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, null=True, blank=True)
```
pxwxnvermx, line 402: nit: Can we make the user foreign key model reference similar on both new models?
i.e using either ConnectUser or settings.AUTH_USER_MODEL for both?

users/email_utils.py
zandre-eng: In the [PR for the view changes](https://github.com/dimagi/connect-id/pull/243/changes#diff-e559aa8a09e3a05a27fa6fcef98aafbd888877551a23cf23ca4ee8e08ab97e47)  we do run a `validate_email` on the email before calling this util. But, agreed good to have some fail code here as well. Will update.

zandre-eng: Validation has been added in the util func in [2de0c62](https://github.com/dimagi/connect-id/pull/242/commits/2de0c62dcecf9f73e6bd2dc23fcd80c5ac3c2459).

users/models.py
zandre-eng: Addressed in [26d44b5](https://github.com/dimagi/connect-id/pull/242/commits/26d44b520667a52d29d79e1f15d56f73cfc8040b).

pxwxnvermx (approved):

PR #248, opened 2026-06-05
handle duplicate key error in message notifications

sravfeyn:
## Technical Summary

https://dimagi.atlassian.net/browse/CI-680
[Sentry](https://dimagi.sentry.io/issues/7511134472/?project=4508576093044736&query=is%3Aunresolved&referrer=issue-stream)

Instead of trying to recreate `Notification` objects, this does a `get_or_create` to fix mode conflict exception

### Safety story

### Automated test coverage

Exists
### Labels & Review

### DIFF
--- a/utils/notification.py
+++ b/utils/notification.py
@@ -3,7 +3,7 @@ from fcm_django.models import FCMDevice
 from firebase_admin import messaging
 from firebase_admin.exceptions import FirebaseError
 
-from messaging.models import Notification
+from messaging.models import Notification, NotificationTypes
 from messaging.serializers import NotificationData
 from users.models import ConnectUser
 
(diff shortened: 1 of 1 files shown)


Charl1996 (approved):

utils/notification.py
```diff
@@ -21,11 +21,21 @@ def send_bulk_notification(message: NotificationData):
...
+            notification, _ = Notification.objects.get_or_create(
+                message_id=message_id,
+                defaults={"user": user, "json": {"title": message.title, "body": message.body, "data": data}},
+            )
+        else:
+            notification = Notification(
+                user=user,
+                json={"title": message.title, "body": message.body, "data": data},
+            )
+            notification.save()
```
pxwxnvermx, line 38: The `get_or_create` here will not save the notification_type field, currently that happens in the overridden `save`  method for the Notification Model. 

I think using the get_or_create is a better approach then overriding the `save` method, so can we update this code to save the `notification_type` field as well and completely remove the `save`  method code from Notification model?

pxwxnvermx: This can be ignored. I checked the code for get_or_create. It implicitly calls save.

pxwxnvermx (approved):

PR #251, opened 2026-06-05
Support Basic auth for email verification and persist verified email to active users

zandre-eng:
## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2434)

This is a release path 3 feature — Experiments & early stage iterations of product area.

The email-OTP endpoints (`send_email_otp`, `verify_email_otp`) previously accepted only OAuth2 Bearer tokens and `ConfigurationSession` tokens, leaving out clients that authenticate with HTTP Basic (`username:password`) — the scheme used by the app's other logged-in-user endpoints (e.g. `change_password`, `update_profile`). Those clients received a `401` (surfaced confusingly as `INVALID_TOKEN`) on the email-verification flow. This PR adds Basic auth to both endpoints, refines where a verified email is persisted based on auth type and account state, and returns a clean `400` when the email collides with the active-user uniqueness constraint instead of a `500`.

- **Authentication:** Added `DeviceBasicAuthentication` to the `authentication_classes` of `send_email_otp` and `verify_email_otp`, alongside the existing `OAuth2Authentication` and `SessionTokenAuthentication`. It is placed first to match the global `DEFAULT_AUTHENTICATION_CLASSES` ordering; this is safe because DRF only short-circuits on a raised error — `BasicAuthentication` returns `None` for non-`Basic` schemes and `OAuth2Authentication` returns `None` on an invalid token, so Bearer/session credentials still reach their correct authenticator.
- **Email persistence on verification:** Reworked the `verify_email_otp` success path so the verified email lands in the right place:
  - **Configuration-session auth + active `ConnectUser` exists for the session's phone number** → the email is written directly to that user.
  - **Configuration-session auth + no active user** → the email is stored on the session (`ConfigurationSession.verified_email`), to be applied later in the flow.
  - **Other auth (OAuth2 / Basic)** → the authenticated user's email is updated directly.
  - The two user-save paths are consolidated into a single save site; `verified_email` is handled inline (with an early return) inside the session branch, so the session-only attribute is never referenced on a non-session auth object.
- **Duplicate-email handling:** The `email_active_user` partial unique constraint (active users, non-empty email) means a verified email already held by another active user would otherwise raise `IntegrityError` → unhandled 500. The save is now guarded to return `400 EMAIL_ALREADY_IN_USE`. A new `EMAIL_ALREADY_IN_USE` error code was added. (No `transaction.atomic()` wrapper is used, consistent with the rest of this module; the app runs in autocommit — `ATOMIC_REQUESTS` is not enabled — so a failed `save()` does not poison a surrounding transaction.)

## Logging and monitoring

No new logging introduced. Existing warning logs for failed email-OTP verification and rate limiting remain unchanged. A drop in `401 INVALID_TOKEN` responses on the email-verification endpoints would indicate Basic-auth clients are now succeeding; `400 EMAIL_ALREADY_IN_USE` responses would surface duplicate-email collisions that previously appeared as 500s.

### Safety story

The authentication change is purely additive — it widens the set of accepted credential *formats* on two endpoints without altering the existing OAuth2/session paths, and the `IsAuthenticated` permission is unchanged, so no endpoint becomes more permissive about *who* may access it. The email-persistence change is scoped to the `verify_email_otp` success path: the active-user lookup filters on `is_active=True` and matches by `phone_number` (unique among active users per the model constraint), so `.first()` is deterministic. The duplicate-email guard relies on the database's `email_active_user` constraint as the source of truth (no TOCTOU window), converting a previously-unhandled 500 into a 400. Validated locally via the test suite below.

### Automated test coverage

`TestSendEmailOtp` / `TestVerifyEmailOtp` in `users/tests/test_views.py`:

- `TestSendEmailOtp::test_basic_auth_flow_creates_device_and_sends_otp` — Basic-authed user hits `send_email_otp`, gets `200`, and a `UserEmailOTPDevice` is created and challenged.
- `TestVerifyEmailOtp::test_basic_auth_success_sets_user_email` — Basic-authed user verifies an OTP and `user.email` is set.
- `TestVerifyEmailOtp::test_session_success_with_active_user_updates_user_email` — session auth with an active user on the session's phone number writes to `user.email` and leaves `session.verified_email` empty.
- `TestVerifyEmailOtp::test_session_success_without_active_user_sets_verified_email` — session auth with no matching active user stores the email on `session.verified_email`.
- `TestVerifyEmailOtp::test_session_success_with_inactive_user_sets_verified_email` — guards the `is_active=True` filter: an inactive user with the same phone is not updated; the email falls back to the session.
- `TestVerifyEmailOtp::test_duplicate_active_email_returns_400` — verifying an email already held by another active user returns `400 EMAIL_ALREADY_IN_USE` and leaves the requester's email unchanged. Runs with `transaction=True` to exercise the autocommit path the view uses in production.

Existing OAuth2 (`user_bearer_client`) and session coverage continues to pass.

### QA Plan

- [ ] With the `EMAIL_OTP_VERIFICATION` flag enabled, authenticate as a registered user via HTTP Basic and call `send_email_otp` then `verify_email_otp` — confirm `200` and that `user.email` is updated.
- [ ] Confirm the existing OAuth2 Bearer-token flow still succeeds on both endpoints (no regression).
- [ ] Using a configuration session whose phone number matches an existing active user, verify an email — confirm the user's `email` is updated and `verified_email` on the session stays empty.
- [ ] Using a configuration session whose phone number has no active user, verify an email — confirm `verified_email` is stored on the session.
- [ ] Attempt to verify an email already in use by another active user — confirm a `400` with `EMAIL_ALREADY_IN_USE` (not a 500), and the requester's email is unchanged.
- [ ] Confirm the `PHONE_NOT_VALIDATED` 403 gate still applies to unvalidated sessions, and that invalid Basic credentials return `401`.

### Labels & Review


users/views.py
```diff
@@ -1047,8 +1047,13 @@ def verify_email_otp(request):
         return JsonResponse({"error_code": ErrorCodes.INCORRECT_OTP}, status=401)
 
     if isinstance(request.auth, ConfigurationSession):
-        request.auth.verified_email = email
-        request.auth.save()
+        user = ConnectUser.objects.filter(phone_number=request.auth.phone_number, is_active=True).first()
+        if user:
+            user.email = email
+            user.save()
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [8704db6](https://github.com/dimagi/connect-id/pull/251/commits/8704db6ba4caabce48c2377a2aff15f75f68a3e1).

OrangeAndGreen (approved):

users/views.py
```diff
@@ -1047,8 +1047,13 @@ def verify_email_otp(request):
         return JsonResponse({"error_code": ErrorCodes.INCORRECT_OTP}, status=401)
 
     if isinstance(request.auth, ConfigurationSession):
-        request.auth.verified_email = email
-        request.auth.save()
+        user = ConnectUser.objects.filter(phone_number=request.auth.phone_number, is_active=True).first()
```
ajeety4, line 1050: nit: checking if multiple records are possible in this query ?

ajeety4 (approved):
LGTM
Just a non blocking query

users/views.py
zandre-eng: We have a DB constraint on (phone_number, is_active=True), so there should only ever be one record in the DB.

PR #252, opened 2026-06-09
Apply A/B-batch dependency upgrades

zandre-eng:
## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2387)
[Link to audit report here](https://gist.github.com/zandre-eng/ff4267a2870dbd84a6961920fd9cbfbd)

Applies the category-A and category-B dependency bumps from the 2026-06-09 dependency audit. **Django is intentionally held at 4.1.7** — the 4.1→5.2 LTS upgrade is its own effort (ticket DEP-1), so every bump here was resolved against the Django 4.1 floor and anything requiring Django ≥4.2 was deliberately deferred rather than force-installed. The lockfile (`requirements/*.txt`) is the source of truth for pins; the `.in` files stay unpinned. Commits are structured so each risk tier can be reverted in isolation.

- **A-batch (`1ed471a`)** — non-breaking minor/patch bumps (celery, boto3, google-auth, google-api-python-client, sentry-sdk, whitenoise, psycopg2, django-otp, DRF, django-environ, zxcvbn + dev: pytest-django, factory-boy, flake8, pre-commit, pip-tools) plus **security-driven transitive pull-forwards** (cryptography 41→48, requests, pyjwt, jwcrypto, sqlparse, protobuf 4→7, idna, pyasn1, urllib3 1→2). No code changes.
- **B-batch, no code changes (`68bb616`)** — low-risk majors: gunicorn 20→26 (clears CVE-2024-1135 / CVE-2024-6827), redis-py 5→8, twilio 7→9, phonenumberslite 8→9; dev pytest 7→9 (clears CVE-2025-71176), pytest-cov 6→7, flake8-isort 6→7.
- **django-phonenumber-field 7→8.0.0 (`748928d`)** — isolated because it required a code change: v8 returns `PhoneNumber` objects from `.values()`, which broke JSON serialization in `GetDemoUsers`; coerced to `str`.
- **black 24→26.5.1 (`4a7fdc0`)** — clears CVE-2026-32274; new stable style reformatted 3 source files; the pre-commit `black` hook rev was bumped to match (it was stale at 23.3.0).
- **Twilio call-surface unit tests (`dc55640`)** — new `utils/tests/test_sms.py` pins the `Client(...)`/`messages.create(...)` surface so a future twilio bump fails CI rather than only at runtime.
- **Deferred to DEP-1 (framework-floor-blocked):** django-anymail (needs Django ≥5.0), django-waffle (≥4.2); partial caps on DRF (3.15.1 — the 3.15.2 CVE-2024-21520 fix needs Django ≥4.2, so that CVE remains open until DEP-1), django-otp (1.5.4), django-phonenumber-field (8.0.0).

## Logging and monitoring

No logging/monitoring changes. `sentry-sdk` was bumped within 2.x (init + `RedisIntegration` unchanged); error reporting continues as before. Success criterion is a clean deploy: workers boot under gunicorn 26 and the Celery worker connects to the redis broker (both verified locally — see Safety story).

### Safety story

Applied incrementally with verification after every commit, against PostgreSQL 16:

- **Full pytest suite green throughout** — 281 passed after each bump commit, **290 passed** after adding the Twilio tests; **pre-commit clean**; `makemigrations --check` reports no drift.
- **Paths the suite can't reach were exercised out-of-band:**
  - **redis-py 8 broker** — ran a real Celery worker with eager mode off; full enqueue→consume→result round-trip through kombu/redis-py 8 succeeded (the suite skips this because `CELERY_TASK_ALWAYS_EAGER=True`). This also clears the kombu `redis<6.5` stale-cap concern.
  - **gunicorn 26** — booted `connectid.wsgi` with the production command (`-w 4`, sync worker); served live HTTP (OIDC discovery 200 — which also exercises the pyjwt/jwcrypto/cryptography signing stack, admin 200, `/` 404), no worker crashes.
- **Changelog review** of every bump on a runtime-integration path, mapped to our exact call sites (urllib3, gunicorn, twilio, redis-py, cryptography, protobuf, firebase-admin, google-api-python-client, google-auth, google-cloud-*) — all NONE/LOW; the Play-Integrity and FCM library surfaces are confirmed unchanged. Full table in `docs/dependency-audit-2026-06-09.md`.
- **Cross-read with `dimagi/commcare-connect#1247`**, whose [staging smoke test](https://github.com/dimagi/commcare-connect/pull/1247) validated Twilio 9.10.9, Sentry 2.x, boto3/S3, Anymail SES, and whitenoise live at versions matching this PR.
- **Inherently bounded:** Django is unchanged, so the framework surface is identical; the only application code change is a one-line serialization coercion in a demo/admin endpoint.
- **Residual:** the live network round-trips to FCM push and Google Play Integrity are not exercised here (need production credentials); their library APIs are changelog-confirmed compatible.

### Automated test coverage

- Existing suite (all pre-existing tests pass on the bumped dependencies), including the OAuth/OIDC, FCM, phone-number, and app-integrity test modules.
- **New `utils/tests/test_sms.py` (9 tests):** `test_get_sms_sender` (parametrized country-code → alpha-sender mapping) and `test_send_sms_invokes_twilio_client` / `test_send_sms_defaults_sender_to_none` — assert `Client(sid, token)` init and the exact `messages.create(body, to, from_, messaging_service_sid)` kwargs, so a breaking Twilio bump is caught in CI.
- The `GetDemoUsers` serialization fix is covered by the existing `TestGetDemoUsers::test_success` (which failed on the raw phonenumber-field 8 bump and passes with the fix).

### QA Plan

- [ ] CI green (pre-commit + pytest) on the PR.
- [ ] After deploy, confirm gunicorn workers boot and the app serves (health/OIDC discovery endpoint responds).
- [ ] Confirm the Celery worker connects to the redis broker and processes a scheduled/async task.
- [ ] Trigger an SMS send (e.g. registration/OTP flow) and confirm delivery via Twilio.
- [ ] Trigger an FCM push (notification/message flow) and confirm delivery on a device.
- [ ] Exercise an app-integrity-protected endpoint to confirm Google Play Integrity verification still succeeds.
- [ ] Confirm a user-photo upload/fetch (S3) works.

### Labels & Review


requirements/requirements.txt
```diff
@@ -2,58 +2,59 @@
...
     # via firebase-admin
-cachetools==5.3.1
-    # via google-auth
-celery==5.4.0
+celery==5.6.3
     # via
     #   -r requirements/requirements.in
     #   django-celery-beat
-certifi==2022.12.7
+certifi==2026.5.20
```
mkangia: just confirming this is a safe jump?

requirements/requirements.txt
```diff
@@ -2,58 +2,59 @@
...
     # via celery
-click-plugins==1.1.1
+click-plugins==1.1.1.2
     # via celery
 click-repl==0.3.0
     # via celery
 cron-descriptor==1.4.5
     # via django-celery-beat
-cryptography==41.0.2
+cryptography==48.0.0
```
mkangia: seems like packages related to crypto got a big jump overall.

requirements/requirements.txt
```diff
@@ -149,7 +165,7 @@ grpcio==1.81.0
     #   grpcio-status
 grpcio-status==1.81.0
     # via google-api-core
-gunicorn==20.1.0
+gunicorn==26.0.0
```
mkangia: that's a big jump.

utils/tests/test_sms.py
mkangia, line 1: curious to learn more about this commit.
Did you do a smoke test and that revealed an issue? We added tests but there was no fix needed?

mkangia (approved):
Some major bumps but guess we will need to just trust the testing done and the tests we have.

requirements/requirements.txt
zandre-eng: This should be quite a safe bump. Claude says the following:

> Yes — certifi is purely the Mozilla CA certificate bundle, no code. The 2022→2026 jump adds newly-trusted roots and removes expired/distrusted ones. The only breakage scenario would be connecting to a server whose CA was pulled from the Mozilla trust store, which means it was already considered untrustworthy. All our endpoints (Twilio, AWS, Firebase, Google, Sentry) use major CAs that are solidly in the bundle.

requirements/requirements.txt
zandre-eng: This should be a safe bump and was smoke tested locally. More context around the smoke test can be found in the linked audit report. Claude's response here:

Yes, safe. We reviewed this one specifically — a few reasons to be confident:

- We don't call it directly — it's transitive via jwcrypto and pyjwt (our JWT/OIDC stack). No cryptography API in our code can break.
- The breaking changes across 42→48 don't touch our path — dropped binary elliptic curves (SECT*; JWT uses NIST prime curves), stricter X.509 CRL validation (not on the JWT path), RSA keygen 1024-bit floor (our keys are ≥2048). None reachable from our code.
- We install prebuilt wheels — the Rust MSRV and OpenSSL bumps in later versions only affect source builds, not the manylinux/abi3 wheels pip installs on Python 3.11.
- It's a security pull-forward — 41.0.2 carried 4 CVEs (including CVE-2024-0727 and CVE-2026-26007); 48.0.0 clears all of them.
- Live-tested — the gunicorn boot smoke test served a valid OIDC discovery response (HTTP 200), which means jwcrypto + pyjwt + cryptography all initialised and signed correctly at 48.0.0.

requirements/requirements.txt
zandre-eng: This should be safe, and was smoke tested locally. Claude's response:

Yes, reviewed and live-tested. Key points:

- Our invocation is unchanged — we run gunicorn connectid.wsgi --bind … --chdir … -w 4 with no config file and the default sync worker. All three flags exist unchanged through 26.0.0, the default sync worker has no breaking changes across any of the majors, and Python 3.11 is supported.
- The only behavioral change that touches a plain Django WSGI deploy is stricter HTTP/1.1 parsing introduced in 22.0 — this is the actual fix for CVE-2024-1135 and CVE-2024-6827 (request smuggling). It rejects malformed requests that 20.x silently accepted. No config required; it's the intended security hardening.
- The eventlet worker was removed in 26.0 — irrelevant since we use the default sync worker.
- Live-tested locally — booted the real connectid.wsgi with the production command, confirmed 4 sync workers came up and served real HTTP responses (OIDC discovery 200, admin 200) with no crashes or tracebacks.

utils/tests/test_sms.py
zandre-eng: > Did you do a smoke test and that revealed an issue?

It was more the other way around. After getting Claude to do a changelog review of the Twilio bump to confirm the changes were safe this test was added as a further verification step. This will also help with future bumps, so we can more easily see if Twilio breaks when we do another upgrade.

mkangia:
Thanks for the responses @zandre-eng

users/views.py
```diff
@@ -674,6 +674,9 @@ def get(self, request, *args, **kwargs):
         )
 
         demo_users = list(demo_phone_devices) + list(demo_connect_users)
+        # phonenumber-field 8 returns PhoneNumber objects from .values(); coerce to str for JSON.
+        for demo_user in demo_users:
+            demo_user["phone_number"] = str(demo_user["phone_number"])
```
hemant10yadav: I think the same needs to be done for FetchUsers as well.

requirements/requirements.txt
```diff
@@ -77,168 +90,206 @@ django-axes[ipware]==6.0.3
...
-django-ipware==5.0.0
+django-ipware==7.0.1
     # via django-axes
 django-oauth-toolkit==2.3.0
     # via -r requirements/requirements.in
-django-otp==1.1.6
+django-otp==1.5.4
     # via -r requirements/requirements.in
-django-phonenumber-field==7.0.2
+django-phonenumber-field==8.0.0
```
hemant10yadav: I noticed in the [changelog](https://github.com/django-phonenumber-field/django-phonenumber-field/releases/tag/8.0.0) that django-phonenumber-field 8.0.0 dropped support for Django 4.1, while we're currently pinned to Django 4.1.7.

zandre-eng: Good catch. Claude checked the PyPi metadata for this upgrade which only listed (Django>=3.2), so this is probably how it fell through. It doesn't look like it affects our runtime as all the unit tests passed, but we should definitely revert this back to a supported version (until the Django upgrade is done).

I also let Claude do a second iteration of changelog review on the bumps based on this and it seems this was the only one that needs to be reverted.

Reverted in [c3784d9](https://github.com/dimagi/connect-id/pull/252/commits/c3784d95ee4043667cd51064671715fc46f5197d).

users/views.py
zandre-eng: The library bump that required this was reverted, so this is no longer necessary.

Reverted in [c3784d9](https://github.com/dimagi/connect-id/pull/252/commits/c3784d95ee4043667cd51064671715fc46f5197d).

hemant10yadav (approved):

zandre-eng:
I resolved a conflict from main which switches PersonalID from pip to uv. I let Claude run a comparison sweep to ensure all dependencies are in the same state as before the merge.

Charl1996 (approved):

PR #253, opened 2026-06-12
Use visible_name as channel_source in FCM payload; propagate channel_name updates

snopoke:
## Technical Summary

Companion to dimagi/open-chat-studio#3620: OCS is moving to sending a stable opaque ID as `channel_source` and the bot's display name as `channel_name`. The mobile app displays whatever arrives under the `"channel_source"` JSON key and ignores `"channel_name"`, so the channel-creation FCM payload now sends `channel.visible_name` there (the sync endpoints already do this) — otherwise the opaque ID would show up in the UI.

Since `get_or_create` only sets `channel_name` on creation, a renamed bot never updated an existing channel; the view now saves a changed `channel_name` on repeat calls. Identity (`channel_source`) stays immutable.

**Note**: The change to use `visible_name` in `channel_source` is a temporary fix until https://dimagi.atlassian.net/browse/CCCT-2509 is resolved.

### Safety story

No data migration; behavior for current callers is unchanged where `channel_name` is absent (`visible_name` falls back to `channel_source`). Name updates only happen when the caller explicitly sends a different `channel_name`.

### Automated test coverage

Tests cover the FCM payload with and without `channel_name`, that re-calling with a new name updates the stored name while returning the same `channel_id`, and that re-calling without a name leaves it unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/messaging/views.py
+++ b/messaging/views.py
@@ -128,12 +128,11 @@ class CreateChannelView(APIView):
         channel_name = data.get("channel_name")
         server = get_current_message_server(request)
         user = get_object_or_404(ConnectUser, username__iexact=connect_id)
-        channel, created = Channel.objects.get_or_create(
-            server=server, connect_user=user, channel_source=channel_source, defaults={"channel_name": channel_name}
+        # only update channel_name if one was provided so an absent name doesn't wipe a stored one
+        defaults = {"channel_name": channel_name} if channel_name else {}
+        channel, created = Channel.objects.update_or_create(
+            server=server, connect_user=user, channel_source=channel_source, defaults=defaults
         )
-        if not created and channel_name and channel.channel_name != channel_name:
-            channel.channel_name = channel_name
-            channel.save(update_fields=["channel_name"])
         response_dict = {"channel_id": str(channel.channel_id), "consent": channel.user_consent}
         if created:
             message = NotificationData(
--- a/messaging/views.py
+++ b/messaging/views.py
@@ -131,6 +131,9 @@ class CreateChannelView(APIView):
         channel, created = Channel.objects.get_or_create(
             server=server, connect_user=user, channel_source=channel_source, defaults={"channel_name": channel_name}
         )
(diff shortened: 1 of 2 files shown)


messaging/views.py
```diff
@@ -131,6 +131,9 @@ def post(self, request, *args, **kwargs):
         channel, created = Channel.objects.get_or_create(
```
snopoke: what about using `create_or_update` instead of `get_or_create`?

snopoke: Fixed in 149bca2. Switched to `update_or_create`; defaults are built conditionally so a request without `channel_name` doesn't wipe a stored name.

*— Claude Code*

messaging/views.py
```diff
@@ -140,7 +142,7 @@ def post(self, request, *args, **kwargs):
                     "body": f"A new messaging channel is available from {channel.visible_name}, press here to view",
                     "key_url": str(server.key_url),
                     "action": CCC_MESSAGE_ACTION,
-                    "channel_source": channel_source,
+                    "channel_source": channel.visible_name,
```
snopoke, line 147:
```suggestion
                    # Mobile uses 'channel_source' for display.
                    # See https://dimagi.atlassian.net/browse/CCCT-2509
                    "channel_source": channel.visible_name,
```

sravfeyn (approved):
Thanks!

messaging/views.py
```diff
@@ -140,7 +142,9 @@ def post(self, request, *args, **kwargs):
                     "body": f"A new messaging channel is available from {channel.visible_name}, press here to view",
                     "key_url": str(server.key_url),
                     "action": CCC_MESSAGE_ACTION,
-                    "channel_source": channel_source,
```
sravfeyn, line 143: Does existing channels/conversations get impacted by this change?

snopoke: no, this only applies when new channels are created.

PR #267, opened 2026-06-26
Management Command to Sync Issued Date for User Credentials

zandre-eng:
## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-1942)

Builds on the [`ze/credential-issued-date`](https://github.com/dimagi/connect-id/pull/266) base branch, which added a per-user `UserCredential.created_at`. Because that column was added with `auto_now_add`, existing rows were stamped with the migration run time rather than the real issue date, so this adds a `sync_usercredential_created_at` management command to backfill the accurate per-user dates from a Connect prod export of `issued_on`. It optionally (`--fix-issuer`) also repoints credentials that were historically mis-attributed to the staging issuer (Connect prod pushed them via the generic/staging client) onto the production issuer, using a split-per-user approach so staging-only holders are preserved.

## Logging and monitoring

The command is a dry run by default and prints a reconciliation summary before any writes — counts of `to_update`, `already_correct`, `no_pid_match`, `pid_unmatched_by_export`, and, with `--fix-issuer`, issuer repoints and the number of staging credentials left orphaned. It is a one-off operational tool and adds no runtime logging.

### Safety story

This is an opt-in management command, not runtime code, so it has no effect until deliberately run. It defaults to a dry run, requires both `--apply` and an interactive `y` confirmation (with `--no-input` for scripted runs), executes within a single transaction, and is idempotent on re-run. Matching is restricted to CONNECT-issued credentials on the join key `(username, type, level, slug)`. `--fix-issuer` creates production-issuer credentials while leaving staging credentials intact and never deletes data — orphaned staging credentials are only reported.

### Automated test coverage

Tests exercise the command end-to-end: loading/parsing the export file, dry-run vs apply, the matching rules (CONNECT-only, int/string opportunity-id coercion), skip/already-correct/no-match/unmatched reporting, bad `issued_on` handling, the confirmation prompt (apply vs abort), input validation, and the optional issuer repointing including the shared-credential split and idempotency.

### QA Plan

### Labels & Review

### DIFF
--- a/users/management/commands/sync_usercredential_created_at.py
+++ b/users/management/commands/sync_usercredential_created_at.py
@@ -1,0 +1,299 @@
+"""
+Reconcile PersonalID UserCredential records against Connect's prod issuance.
+
+Two things, both driven by a JSON export of Connect prod UserCredential.issued_on
+(format described below):
+
+1. created_at backfill (always): UserCredential.created_at (migration 0032) was added
+   with auto_now_add, so existing rows hold the migration run time rather than the date
+   the credential was issued to the user. Credential.created_at is not a substitute (it is
+   per-credential, shared across holders and environments). The per-user source of truth is
+   Connect's UserCredential.issued_on, which this command writes onto created_at.
+
+2. issuer fix (--fix-issuer, opt-in): because Connect prod historically pushed credentials
+   using the generic (staging) client, prod-issued credentials are attributed to the
+   CONNECT/staging IssuingAuthority. For each prod-matched UserCredential this repoints it to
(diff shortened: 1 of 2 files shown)


Charl1996 (dismissed):
Scanned through it. Looks functionally correct.

It seems like this command is here to stay, or will we run it once and then remove it?

zandre-eng:
> It seems like this command is here to stay, or will we run it once and then remove it?

@Charl1996 I'm opted to remove this command once it's been run, as new credentials going forward should have the correct issued date given automatically.

sravfeyn (dismissed):
LGTM

Charl1996 (approved):

PR #270, opened 2026-07-15
Batch FCM sends in bulk notification endpoint

zandre-eng:
## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CI-788)

The daily 07:00 UTC response-time spike traced to `/messaging/send_bulk/`: commcare-connect's `send_notification_inactive_users` beat task posts the whole recipient batch in one request, and the endpoint then sent one blocking FCM request per recipient, sequentially, on a single Gunicorn sync worker — pinning that worker for the full batch while other traffic queued behind it. This change dispatches all pushes via `messaging.send_each` in batches of `MAX_MESSAGES_PER_BATCH` (concurrent within each batch) and persists the notifications in a single transaction.

## Logging and monitoring

No new logging is added. The effect is observable through existing request-duration metrics for `POST /messaging/send_bulk/` and RDS write load around 07:00 UTC, both of which should drop after rollout.

### Safety story

The change is confined to the notification-sending path. The response contract (per-message `all_success` and per-user `status`) is unchanged and asserted by the existing tests, so callers, none of which currently consume the response body, are unaffected. Notification persistence now happens in a single transaction (atomic per bulk request), and bad-token device deactivation behavior is preserved via fcm-django's existing helper. The change is a straight revert away if needed.

### Automated test coverage

Tests cover the bulk endpoint across success, no-active-device / inactive / missing-user (→ `deactivated`), unregistered-token and generic FCM-error mapping, duplicate-username dedup with stable ordering, and both notification-creation paths (plain and messaging-type). The exact FCM message payloads and the full response body are asserted to confirm the contract is unchanged. (Not yet covered: batches exceeding `MAX_MESSAGES_PER_BATCH`, and mixed success/failure result mapping within a single multi-item batch.)

### QA Plan

### Labels & Review

### DIFF
--- a/messaging/serializers.py
+++ b/messaging/serializers.py
@@ -36,9 +36,13 @@ class SingleMessageSerializer(serializers.Serializer):
     fcm_options = serializers.DictField(required=False, default={})
 
     def create(self, validated_data):
+        validated_data = dict(validated_data)
         username = validated_data.pop("username", None)
         if username:
-            validated_data["usernames"] = [username]
+            usernames = list(validated_data.get("usernames") or [])
+            if username not in usernames:
+                usernames.append(username)
+            validated_data["usernames"] = usernames
         return NotificationData(**validated_data)
 
 
@@ -55,7 +59,8 @@ class BulkMessageSerializer(serializers.Serializer):
         return messages
 
     def create(self, validated_data):
-        return [NotificationData(**message) for message in validated_data["messages"]]
+        child = self.fields["messages"].child
+        return [child.create(message) for message in validated_data["messages"]]
 
 
 class MessageSerializer(serializers.ModelSerializer):
--- a/messaging/serializers.py
+++ b/messaging/serializers.py
@@ -5,6 +5,8 @@ from rest_framework import serializers
 from messaging.models import Message, Notification, NotificationTypes
 
 CCC_MESSAGE_ACTION = "ccc_message"
+MAX_BULK_MESSAGES = 5000
+MAX_BULK_RECIPIENTS = 1000
 
 
 @dataclasses.dataclass
@@ -14,6 +16,15 @@ class NotificationData:
     body: str = None
     data: dict = None
     fcm_options: dict = dataclasses.field(default_factory=lambda: {})
+
+
+def _recipients(message: dict) -> list[str]:
+    """Recipient usernames of a validated single-message payload, from either the singular or plural key."""
+    usernames = list(message.get("usernames") or [])
+    username = message.get("username")
+    if username:
+        usernames.append(username)
+    return usernames
 
 
 class SingleMessageSerializer(serializers.Serializer):
@@ -32,7 +43,16 @@ class SingleMessageSerializer(serializers.Serializer):
 
 
 class BulkMessageSerializer(serializers.Serializer):
-    messages = serializers.ListField(child=SingleMessageSerializer())
+    messages = serializers.ListField(child=SingleMessageSerializer(), max_length=MAX_BULK_MESSAGES)
+
+    def validate_messages(self, messages):
+        recipients = sum(len(set(_recipients(message))) for message in messages)
+        if recipients > MAX_BULK_RECIPIENTS:
+            raise serializers.ValidationError(
+                f"Too many recipients: {recipients} (maximum {MAX_BULK_RECIPIENTS} per request). "
+                f"Split the messages across multiple requests."
+            )
+        return messages
 
     def create(self, validated_data):
         return [NotificationData(**message) for message in validated_data["messages"]]
--- a/messaging/tests.py
+++ b/messaging/tests.py
@@ -41,7 +41,7 @@ def server():
 def test_send_message(authed_client, fcm_device):
     url = reverse("messaging:send_message")
 
-    with mock.patch("fcm_django.models.messaging.send", wraps=_fake_send) as mock_send_message:
+    with mock.patch("firebase_admin.messaging.send_each", wraps=_fake_send_each) as mock_send_message:
         response = authed_client.post(
             url,
             data=json.dumps(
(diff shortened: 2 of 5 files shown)


utils/notification.py
```diff
@@ -1,68 +1,112 @@
...
+    else:
+        notification = Notification(user=user, json=payload)
+        notification.save()
+    return notification
+
+
+def _send_fcm_messages(fcm_messages: list[messaging.Message]) -> list[messaging.SendResponse]:
+    send_responses: list[messaging.SendResponse] = []
+    for batch in batched(fcm_messages, MAX_MESSAGES_PER_BATCH):
+        send_responses.extend(messaging.send_each(batch).responses)
```
ajeety4: Should we also catch exceptions as stated [here](https://github.com/xtrinch/fcm-django#sending-messages-in-bulk) in the end for the `send_each` ?

messaging/views.py
```diff
@@ -107,13 +107,8 @@ def post(self, request, *args, **kwargs):
...
         messages = serializer.save()
 
-        global_all_success = True
-        results = []
-        for message in messages:
-            message_result = send_bulk_notification(message)
-            results.append(message_result)
-            if not message_result["all_success"]:
-                global_all_success = False
+        results = send_bulk_notifications(messages)
```
ajeety4, line 110: Since this is still a sync request and bounded at 30s on both ends - commcare-connect's client timeout, and gunicorn's default --timeout (docker/start sets none, so the worker gets killed mid-batch) , should we cap the number of messages per request? 

Happy to leave it for a follow-up.

ajeety4 (commented):
LGTM
Just a query around error handling.

utils/notification.py
zandre-eng: I think this would be good to do. Added exception handling in [ecc1d51](https://github.com/dimagi/connect-id/pull/270/commits/ecc1d51acbc19454ed746660c620b4654132cd90).

messaging/views.py
zandre-eng: Good idea. I've added limits in [9853ca7](https://github.com/dimagi/connect-id/pull/270/commits/9853ca78c1913c4b4eb5acf19105a1e4226fad20).

messaging/tests.py
```diff
@@ -150,8 +157,45 @@ def test_send_message_bulk(authed_client, fcm_device):
         ]
 
 
-def _fake_send(messages, **kwargs):
-    return messaging.BatchResponse([messaging.SendResponse({"name": f"message_id_{random.randint(10, 40)}"}, None)])
+def _bulk_messages(recipients, num_messages=1, repeat=1):
+    """`num_messages` payloads splitting `recipients` distinct usernames, each username listed `repeat` times."""
+    usernames = [f"user-{index}" for index in range(recipients)]
+    chunks = list(batched(usernames, -(-recipients // num_messages))) if recipients else [[]] * num_messages
+    return [{"usernames": list(chunk) * repeat, "body": "test message"} for chunk in chunks]
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [d897eff](https://github.com/dimagi/connect-id/pull/270/commits/d897eff004748a24a4a955de26723ffc5f8e4ddc).

messaging/serializers.py
```diff
@@ -16,6 +18,15 @@ class NotificationData:
     fcm_options: dict = dataclasses.field(default_factory=lambda: {})
 
 
+def _recipients(message: dict) -> list[str]:
+    """Recipient usernames of a validated single-message payload, from either the singular or plural key."""
+    usernames = list(message.get("usernames") or [])
+    username = message.get("username")
+    if username:
+        usernames.append(username)
+    return usernames
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [d897eff](https://github.com/dimagi/connect-id/pull/270/commits/d897eff004748a24a4a955de26723ffc5f8e4ddc).

messaging/serializers.py
```diff
@@ -32,7 +43,16 @@ def create(self, validated_data):
 
 
 class BulkMessageSerializer(serializers.Serializer):
-    messages = serializers.ListField(child=SingleMessageSerializer())
+    messages = serializers.ListField(child=SingleMessageSerializer(), max_length=MAX_BULK_MESSAGES)
+
+    def validate_messages(self, messages):
+        recipients = sum(len(set(_recipients(message))) for message in messages)
```
ajeety4, line 53: Should this count unique recipients across the whole request? Right now the same recipient in multiple messages will be counted multiple times. Is that the intended behaviour?

ajeety4 (approved):
Thanks for the changes.
Just a minor query.

messaging/serializers.py
zandre-eng: Good question. Yes, this is intended. The cap limits how many notifications we send, not how many people we send to. Within a single message we do dedupe, since repeats there collapse into one notification.

Counting each person only once request-wide would mean 1000 users spread across 5000 messages passes the validation check but will send millions of notifications.

messaging/serializers.py
```diff
@@ -5,6 +5,8 @@
 from messaging.models import Message, Notification, NotificationTypes
 
 CCC_MESSAGE_ACTION = "ccc_message"
+MAX_BULK_MESSAGES = 5000
+MAX_BULK_RECIPIENTS = 1000
```
Charl1996, line 9: Where did this number come from?

Charl1996 (approved):
Did a high level pass and it seems functionally correct. Thanks for the improvement

messaging/serializers.py
zandre-eng: Nowhere specific. Claude suggested this value as a safe starting point which can later be taken up if needed. The important const is the `MAX_BULK_RECIPIENTS` which limits the request to FCM pushes. Happy to change this value if you think it's off.

PR #276, opened 2026-07-29
Upgrade Django to 4.2 and django-oauth-toolkit to 3.3.0

hemant10yadav:
## Technical Summary

Ticket: [CCCT-2496](https://dimagi.atlassian.net/browse/CCCT-2496)

- Upgraded Django from 4.1.7 to 4.2.30, required by django-oauth-toolkit 3.3.0
- Upgraded django-oauth-toolkit from 2.3.0 to 3.3.0
- Added a test suite pinning current OAuth2/OIDC behavior (grants, revocation, introspection, discovery, userinfo) before the upgrade, so any behavior change from either library would show up in CI

## Logging and monitoring

No new logging or monitoring needed for this change.

### Safety story

Wrote tests covering the existing OAuth2/OIDC behavior first, then upgraded both packages and confirmed the full suite still passes against the new versions. Also ran the app locally to sanity-check login and token flows.

### Automated test coverage

Added `users/tests/test_oauth.py` covering password/client_credentials/refresh grants, token revocation, introspection, and OIDC discovery/userinfo; all passing.

### QA Plan

No QA ticket needed. Tested manually on my local env using Postman, covering the OAuth flows and basic functionality.

### Labels & Review

[CCCT-2496]: https://dimagi.atlassian.net/browse/CCCT-2496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/users/tests/test_oauth.py
+++ b/users/tests/test_oauth.py
@@ -1,8 +1,12 @@
+import base64
 from datetime import timedelta
 from unittest.mock import MagicMock
 
 import pytest
+from django.urls import reverse
 from django.utils.timezone import now
+from oauth2_provider.models import AccessToken, Application, RefreshToken
+from rest_framework.test import APIClient
 
 from users.const import ErrorCodes
 from users.factories import UserDeviceInfoFactory, UserFactory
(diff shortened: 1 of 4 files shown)


users/tests/test_oauth.py
```diff
@@ -79,3 +108,236 @@ def test_failed_auth_old_access_returns_false(self):
...
+            HTTP_AUTHORIZATION=_basic_auth_header(password_grant_app.client_id, password_grant_app.raw_client_secret),
+        )
+        assert revoke_response.status_code == 200
+
+        api_client = APIClient()
+        api_client.credentials(HTTP_AUTHORIZATION=f"Bearer {issued['access_token']}")
+        response = api_client.get(reverse("oauth2_provider:user-info"))
+        assert response.status_code in (401, 403)
+
+    def test_deactivate_account_revokes_all_user_tokens(self, oauth_app, user):
```
ajeety4: nit: This feels more like a token revocation test than a deactivation test. Since it's only verifying the revoke logic (which happens during account deactivation), would it make sense to rename the test to better reflect what it's actually testing?
Maybe something like `test_token_revocation_used_by_deactivate_account`

ajeety4 (approved):
LGTM. 
Appreciate the test cases added.

users/tests/test_oauth.py
hemant10yadav: Done

pxwxnvermx (approved):

PR #277, opened 2026-07-30
Upgrade Django to 4.2 LTS and fcm-django to 3.2

zandre-eng:
## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2497)

This is a release path 1 feature — Improvements to existing features & quick wins

Two coupled upgrades: Django 4.1.7 → 4.2.27 (LTS) and fcm-django 2.2.1 → 3.2.0. They are
coupled because fcm-django 3.2 requires `Django>=4.2`, so the framework bump is a
prerequisite rather than an independent choice. They are nonetheless split into separate
commits, and each was verified green on its own, so either can be reverted without the
other — Django 4.2 passes the full suite with fcm-django still at 2.2.1.

Worth flagging one correction to the ticket's premise: it describes fcm-django 3.x as
dropping the legacy FCM API in favour of Firebase HTTP v1. That migration actually landed
in fcm-django **2.0**, so this repo was already on the v1 API (`messaging.send_each`,
`credentials.Certificate` + `initialize_app`). There was no message-construction rewrite
or credential reconciliation to do, and every fcm-django API this codebase touches
(`MAX_MESSAGES_PER_BATCH`, `DeviceType`, `FCMDevice`,
`deactivate_devices_with_error_results`) is signature-identical in 3.2. The real work was
the forced Django bump.

## Logging and monitoring

A post-deploy verification script will be run in the container after release. It asserts the deployed
package versions, that no migrations are pending, that the Firebase app initialises against
the HTTP v1 endpoint, that the service account mints an access token, that the send path is
reachable (FCM rejects a deliberately fake token — nothing is delivered to any device), that
Celery workers are up with the FCM task registered, and that email still resolves to the
anymail SES backend.

Two things to watch after release:

- **Sentry**, for the first hour: `FirebaseError` / `InvalidArgumentError` (send path),
  `NotRegistered` (Celery), `DisallowedHost` (AllowCIDRMiddleware), `FieldError` from
  `/admin/`, and `NumberParseException` (the phonenumber-field bump).
- **`FCMDevice` active/inactive counts**, recorded at deploy and re-checked a week later.
  fcm-django 3.x no longer deactivates a device on every `INVALID_ARGUMENT`, so inactive-device
  growth should slow. If it flatlines entirely while `INVALID_ARGUMENT` rates climb, the
  change went further than intended and deactivating on specific causes should be revisited.

### Safety story

Confidence came from several passes rather than one. Every backwards-incompatible change and
deprecation in the Django 4.2 notes was checked against this codebase and none apply —
including the one 4.2 change that can silently drop field writes, `update_or_create()`
passing `update_fields` into a custom `save()`, which was traced on both sides and ruled out.
All 27 intervening 4.2.x patch releases were reviewed; the only two actionable items don't
apply (the multiple-file-upload widget change needs Django forms, of which this repo has
none, and the PostgreSQL index regression only affects indexes built under 4.2.0–4.2.5, a
window this deployment skips entirely). The suite also passes with Django's
`RemovedInDjango50Warning` and `RemovedInDjango51Warning` promoted to errors.

### Automated test coverage

Coverage was added for three paths that previously had none: the Firebase credential
construction and app initialisation, `AllowCIDRMiddleware`'s host admission (first in the
middleware stack, and its library does not declare Django 4.2 support), and the `fetch_users`
endpoint. The FCM send tests now pin all three device-deactivation outcomes across Firebase
error causes, locking in fcm-django 3.x's changed behaviour, and a guard asserts the
phone-number serialisation contract so a future `django-phonenumber-field` bump fails in CI
rather than returning 500s in production. Each of the six commits was checked out, synced to
its own lockfile and run independently — 346 tests passing at branch head, up from 329 on
`main`.

### QA Plan

### Labels & Review


connectid/tests.py
pxwxnvermx, line 1: Can these tests be removed? I don't think they offer much value as they are testing an external library for validations that run on a config.

connectid/firebase.py
pxwxnvermx, line 1: These helpers are not contributing much other than just unlocking tests for an external library by extracting firebase env loading into functions. This can also be reverted if the related test file is removed.

pyproject.toml
```diff
@@ -5,18 +5,18 @@ requires-python = "==3.11.*"
 dependencies = [
     "boto3==1.43.25",
     "celery==5.6.3",
-    "django==4.1.7",
+    "django==4.2.27",
```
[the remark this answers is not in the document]
zandre-eng: A django upgrade to 5.2 will be done in a separate PR, so not relevant here. The Django upgrade in this PR was just so that a compatible version was used for the `fcm-django` upgrade.

utils/tests/test_middleware.py
pxwxnvermx, line 1: I think this file is also testing an external library and can be removed.
django-allow-cidr does declare support for Django 4.2 and 5.2 in the latest [tagged release](https://github.com/mozmeao/django-allow-cidr/releases/tag/0.8.0).

pxwxnvermx (commented):
Couple of comments about the testing included for external libraries in this PR. Rest of the changes look good.

zandre-eng:
@pxwxnvermx I have reverted the tests and extracted helper functions as called out by your feedback. Please could you give another pass review?

CLAUDE.md
```diff
@@ -57,7 +57,7 @@ deploy/             # Kamal deployment config
 
 ## Tech Stack
 
-- **Python 3.11**, Django 4.1, Django REST Framework
+- **Python 3.11**, Django 4.2 (LTS), Django REST Framework
```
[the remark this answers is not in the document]
zandre-eng: This should get changed with an upcoming PR to upgrade Django to 5.2, so won't be addressing that here.

pxwxnvermx (approved):

hemant10yadav (approved):
LGTM

PR #281, opened 2026-08-06
Always Allow OTP Fallback in Personal ID signups

shubham1g5:
https://dimagi.atlassian.net/browse/CI-877

### Safety story

- Modified existing tests to check for truthyness of the `otp_fallback` parameter. 

- This change exposes Twilio to all regions as any user can utilize this fallback in case of Firebase Auth failures on mobile. It doesn't seem like we are stop-gating the Twilio request to first check for regions it's available in for the platform.  Although it seems like an otp request to Twilio in a non-supported region should fail with an exception and abort the request with a `500`.  So there doesn't seem to be any real risks here of exposing users to Twilio other than they will fail the requests which seems like the right outcome in this scenario. 

### QA Plan

None 

### Labels & Review

### DIFF
--- a/users/tests/test_views.py
+++ b/users/tests/test_views.py
@@ -1492,7 +1492,7 @@ class TestStartConfigurationView:
 
         sms_method = response.json().get("sms_method")
         assert sms_method == SMSMethods.FIREBASE
-        assert response.json().get("otp_fallback")
+        assert response.json().get("otp_fallback") is True
 
     @skip_app_integrity_check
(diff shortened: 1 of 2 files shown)


Charl1996 (approved):
The code changes looks fine, but curious what @calellowitz 's inputs are here since this will affect the Twilio bill (not sure how much we care about that).

shubham1g5:
Offlined with Cal and he was good with the change.

